PR #10, opened 2022-10-15
[improve][docs] Fix some typos

andrasbeni:


docs/Scalable Metadata Service.md
```diff
@@ -55,7 +55,7 @@ Only the data path needs to be highly optimized for performance. The control pat
 ## Components
 
 - **Client** → Fetches the list of shards assignments so that it will be able to connect to the right storage pod, the one that is leading a particular shard
-- **Storage pod** → Shards are assigned to storage pods, which can be either &quot;leader&quot; or &quot;follower&quot; on them. Storage pod _do not_ perform health/liveness checks against each others.
+- **Storage pod** → Shards are assigned to storage pods, which can be either &quot;leader&quot; or &quot;follower&quot; on them. Storage pod _do not_ perform health/aliveness checks against each others.
```
teabot: I think Kubernetes uses the term 'liveness' so perhaps best to keep as is: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

teabot: Perhaps undo these XML entity references and use `'` instead.

docs/Scalable Metadata Service.md
```diff
@@ -259,4 +259,4 @@ On the leader side, serving the snapshot is implemented by having a &quot;consis
 
 ## Changing the replication factor
 
-It is possible to dynamically update update the replication factor for the cluster. The operator will apply the change by closing the current epoch for a shard and informing the leader of the new replication factor. The followers will be either started or stopped, depending on whether it&#39;s increasing or decreasing the replication factor. The leader will then use the new size when deciding when an entry is fully committed.
+It is possible to dynamically update the replication factor for the cluster. The operator will apply the change by closing the current epoch for a shard and informing the leader of the new replication factor. The followers will be either started or stopped, depending on whether it&#39;s increasing or decreasing the replication factor. The leader will then use the new size when deciding when an entry is fully committed.
```
teabot:
```suggestion
It is possible to dynamically update the replication factor for the cluster. The operator will apply the change by closing the current epoch for a shard and informing the leader of the new replication factor. The followers will be either started or stopped, depending on whether it's increasing or decreasing the replication factor. The leader will then use the new size when deciding when an entry is fully committed.
```

docs/Scalable Metadata Service.md
teabot:
```suggestion
- **Storage pod** → Shards are assigned to storage pods, which can be either 'leader' or 'follower'. Storage pods _do not_ perform health/liveness checks against each other.
```

docs/Scalable Metadata Service.md
```diff
@@ -186,7 +186,7 @@ Any write operation will be appended to the local WAL. Once enough followers, se
 
 Followers are meant to be very close the leader, tailing the leader WAL and applying all the changes, up to the LastAddConfirmed (same concept as in BookKeeper), which represents the latest entry that we know for sure that it was fully committed.
```
teabot:
```suggestion
Followers are meant to be very close the leader, tailing the leader WAL and applying all the changes, up to the `LastAddConfirmed` (same concept as in BookKeeper), which represents the latest entry that we know for sure that it was fully committed.
```

PR #11, opened 2022-10-17
Use go install and pin versions of protobuf/grpc

nahguam:


.github/workflows/pr_build_and_test.yaml
```diff
@@ -29,11 +29,11 @@ jobs:
 
       - name: Installing protoc-gen-go
         run: |
-          go get github.com/golang/protobuf/protoc-gen-go
-          go get google.golang.org/grpc/cmd/protoc-gen-go-grpc  
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
```
asafm, line 32: Isn't there a module / package manager fo that?

PR #13, opened 2022-10-21
[WIP][Do-not-merge][draft] Replication protocol in worker

andrasbeni:


coordination/coordination.proto
```diff
@@ -0,0 +1,138 @@
+syntax = "proto3";
+
+option go_package = "github.com/streamnative/oxia/coordination";
+
+package proto;
+
+service OxiaCoordination {
+  //controller -> node
```
merlimat, line 8: Given that there are different types of calls `controller -> node` and `node -> node`, it might be better to split this service into 2. It will help to separate and clarify who's calling which method.

merlimat: For naming it could be something like: 
 * `OxiaLogReplication` (leader -> follower)
 * `OxiaControl` (controller -> node)

coordination/coordination.proto
```diff
@@ -0,0 +1,138 @@
...
+
+message ServerAddress {
+  string internal_url = 1;
+  string public_url = 2;
+}
+
+message EntryId {
+  uint64 epoch = 1;
+  uint64 offset = 2;
+}
```
merlimat, line 34: Also, I'd like to change a bit the offset to keep the offset as the unique entry id here. 

instead of having (epoch0, offset(0 -> N) ), we can have the offset to be monotonically increasing. It still is associated with an epoch, though, just it would be enough to specify the offset. 

An epoch becomes just the interval between 2 offsets.

server/tidwall_log.go
```diff
@@ -0,0 +1,227 @@
+package server
+
+import (
```
merlimat, line 3: Same here, let split out the wal implementation

server/BoltKV.go
```diff
@@ -0,0 +1,87 @@
...
+type boltKV struct {
+	store *bolt.DB
+	shard string
+}
+
+func (b *boltKV) Close() error {
+	return b.store.Close()
+}
+
+func (b *boltKV) Apply(op *proto.PutOp, timestamp uint64) (KVEntry, error) {
```
merlimat, line 19: Let's try to break this down into multiple PRs

coordination/coordination.proto
andrasbeni: @merlimat Do you propose to get rid of the epoch in `EntryId`s? I'm pretty sure it could lead to log divergence (and I'll try to prove this only if you really want to get rid of the epoch). 

I agree that within the committed prefix, the offset would be a unique identifier on its own. But where would it be enough to specify just the offset without the epoch?

merlimat: We don't have to change it now. 

The 2 approaches are essentially equivalent: 
 1. `(epoch, entryId)` (where entryId restarts from 0 each new epoch)
 2. `(entryId)` where it monotonically increases. It can still be mapped back to a specific epoch
 
 When a new epoch is advanced, the log needs anyway to be truncated at the entry specified by the controller.

andrasbeni: OK, I will not change this now. What I was thinking of is definitely an edge case (in the example below I will use monotonically increasing offsets) :
Let's say worker `A` is leader and goes down when its head index is `{e:1, o:7}` and the commit index is `{e:1,o:6}`. New leader (worker `B`) is elected, nodes truncate to `{e:1,o:6}` and proceed to commit `{e:2,o:7}`...`{e:2,o:9}`. 
`B` goes down and `A` comes back to participate in the the next leader election. Newly elected leader `C` will get `A`'s head index (`{e:1,o:7}`), sees that in its log the highest entry of the same epoch is `{e:1,o:6}`, so tells `A` to truncate to that.
If epochs are not used in entry identification, how will `C` know to tell `A` to truncate?

coordination/coordination.proto
andrasbeni: Would you like separate .proto files for the separate services? Maybe even separate directories?

merlimat: Either way. Single file works, though we can split into 2 `service` definitions.

coordination/coordination.proto
merlimat: It has to be recorded at what entryId a particular epoch closed, so that everyone can truncate the log there.

PR #17, opened 2022-11-07
Fixed running of CI tests

merlimat:
Tests are not currently running when PRs are created

### DIFF
--- a/.github/workflows/pr_build_and_test.yaml
+++ b/.github/workflows/pr_build_and_test.yaml
@@ -9,37 +9,37 @@ on: [pull_request]
 #
 
 jobs:
-  unit-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_build_and_test.yaml
```diff
@@ -1,4 +1,4 @@
-name: Build & Test
+name: 'Build & Test'
```
nahguam: Shouldn't need this right? It was working before, and I see recent builds still

merlimat: Yes, it doesn't really matter.. It works for merged into `main`, just not PRs: https://github.com/streamnative/oxia/actions

PR #19, opened 2022-11-08
Added DB interface and impl

merlimat:
This PR is stacked on top of #18 

Adding interface for `DB` which handles all the logic for keys `Stat`, versioning and applying batches of operations on the database.


server/kv/db.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (d *db) ProcessBatch(b *proto.BatchRequest) (*proto.BatchResponse, error) {
+	res := &proto.BatchResponse{}
+
+	batch := d.kv.NewWriteBatch()
+	now := uint64(time.Now().UnixMilli())
+
+	for _, putReq := range b.GetPut() {
+		se, err := checkExpectedVersion(batch, putReq.Key, putReq.ExpectedVersion)
```
andrasbeni:
```suggestion
		response, err := applyPut(putReq, batch, now)
		if err != nil {
			return nil, err
		}
		res.Put = append(res.Put, response)
		
```
I'd prefer the individual apply operations refactored into their own methods.

merlimat: Good point. Changed that.

PR #20, opened 2022-11-08
[Feature] Client proto service

nahguam:
First cut of the client protobuf service for Milestone 1

### DIFF
--- a/client/client.proto
+++ b/client/client.proto
@@ -65,22 +65,14 @@ message ShardAssignmentsResponse {
  * The assignment of a shard to a server.
  */
 message ShardAssignment {
-  // The shard
-  Shard shard = 1;
-  // The shard leader, e.g. `host:port`
-  string leader = 2;
-}
-
-/**
- * A shard with the range of hashes that it can contain.
- */
-message Shard {
   // The shard id
-  uint32 id = 1;
+  uint32 shard_id = 1;
   // The minimum inclusive hash that the shard can contain
   uint32 min_hash_inclusive = 2;
   // The maximum exclusive hash that the shard can contain
   uint32 max_hash_exclusive = 3;
+  // The shard leader, e.g. `host:port`
+  string leader = 4;
 }
 
 /**
--- a/docs/proto/client.md
+++ b/docs/proto/client.md
@@ -17,7 +17,6 @@
     - [GetResponse](#io-streamnative-oxia-client-GetResponse)
     - [PutRequest](#io-streamnative-oxia-client-PutRequest)
     - [PutResponse](#io-streamnative-oxia-client-PutResponse)
-    - [Shard](#io-streamnative-oxia-client-Shard)
     - [ShardAssignment](#io-streamnative-oxia-client-ShardAssignment)
     - [ShardAssignmentsRequest](#io-streamnative-oxia-client-ShardAssignmentsRequest)
     - [ShardAssignmentsResponse](#io-streamnative-oxia-client-ShardAssignmentsResponse)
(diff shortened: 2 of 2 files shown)


docs/proto/client.md
```diff
@@ -0,0 +1,394 @@
+# Protocol Documentation
```
andrasbeni, line 1: If I understand right, this is the output of `make proto_doc`. Do we need to add this file to git given that it can be regenerated anytime?

nahguam: Correct. I thought it would be useful to publish

andrasbeni: I agree with publishing it. It has a place in the documentation. I just don't think we need to keep it in git. But I don't feel strongly against it.

client/client.proto
```diff
@@ -0,0 +1,267 @@
...
+  // The message associated with the error
+  string message = 2;
+}
+
+/**
+ * Represents all the possible errors and that can be encountered.
+ */
+enum ErrorType {
+  // Unspecified error
+  ERROR_TYPE_UNSPECIFIED = 0;
```
merlimat, line 264: We should keep `0 == OK` for errors, so that default is no error

client/client.proto
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * One or more items in a batch may fail. For this scenario an error must be
+ * returned for a single item so that the whole batch is not failed and can
+ * continue being processed.
+ */
+message Error {
+  // The specific error type
+  ErrorType type = 1;
+  // The message associated with the error
+  string message = 2;
```
merlimat, line 256: For individual keys errors, there should be no needed extra custom message, because all the errors are pre-determined. If there are unexpected errors, we should always fail the entire batch operation, with a grpc error

client/client.proto
```diff
@@ -0,0 +1,267 @@
...
+ * A put request. Persists the specified key and payload
+ */
+message PutRequest {
+  // The key
+  string key = 1;
+  // The payload
+  bytes payload = 2;
+  // An optional expected version. The put will fail if the actual version
+  // does not match
+  optional uint64 expected_version = 3;
```
merlimat, line 128: better to use `int64` for version because typically `-1` is use to expect the key to not exist

client/client.proto
merlimat: I'd just keep a simple `Error` or `Status` enum.

client/client.proto
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/**
+ * The response to a put request.
+ */
+message PutResponse {
+  // The key
+  string key = 1;
```
merlimat, line 136: Key should not be needed here, since the responses are positional

client/client.proto
```diff
@@ -0,0 +1,267 @@
...
+  oneof result {
```
merlimat, line 137: `oneof` is generating a bit of funky to use code, at least in go. 

Since all values are anyway optional in proto3, I'd just use empty stat and error != 0 to signal a failure.

client/client.proto
nahguam: The linter requires `UNSPECIFIED` as the first item = 0;

client/client.proto
nahguam: I'll change that

client/client.proto
nahguam: 👍

client/client.proto
merlimat: No worries. I'm doing the changes in #19, since I'm using these definition for the DB interface

client/client.proto
merlimat: Oh, let me check that

client/client.proto
nahguam: They are but you get extra presence information in the generated code and I think it's a clearer api with the oneof

client/client.proto
merlimat: I don't necessarily agree with the linter here :) 

Also the generated enum values are already scoped with the enum type name, so I don't think it's necessary to force the type name twice..

client/client.proto
nahguam: No, it's not needed for the put. It's not required for the delete or get either, but is for the delete_range and get_range. I left them in for less message bloat and consistency.

PR #23, opened 2022-11-09
Add DB GetRange DeleteRange operations

merlimat:
Add DB GetRange DeleteRange operations


server/kv/db.go
```diff
@@ -59,6 +59,14 @@ func (d *db) ProcessBatch(b *proto.BatchRequest) (*proto.BatchResponse, error) {
 		}
 	}
 
+	for _, delRangeReq := range b.DeleteRanges {
+		if dr, err := applyDeleteRange(batch, delRangeReq); err != nil {
```
nahguam, line 63: I like this shorthand syntax :)

PR #26, opened 2022-11-09
Client interface and memory impl & tests

nahguam:


client/client.go
```diff
@@ -0,0 +1,44 @@
+package client
```
merlimat: I think we may need to put the client API in `oxia` package (and directory). 

This is because otherwise the users will have to import `github.com/streamnative/oxia/client` and then use it as `client.OxiaClient`, though `client` package name will possibly generate name conflicts.

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"io"
+)
+
+const (
+	AnyVersion int64 = -1
```
merlimat: In many cases, `-1` is used to indicate a `NonExisting` key

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+	NewVersion int64 = 0
+)
+
+var (
+	errorKeyNotFound = errors.New("Key not found")
```
merlimat:
```suggestion
	ErrorKeyNotFound = errors.New("Key not found")
```

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+	errorBadVersion  = errors.New("Bad version")
+)
+
+func ErrorKeyNotFound() error {
```
merlimat: no need to export it through a function, just make the variable public (by using first letter uppercase)

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type Stat struct {
+	Version           int64
+	CreatedTimestamp  uint64
+	ModifiedTimestamp uint64
+}
+
+type Value struct {
+	Payload *[]byte
```
merlimat:
```suggestion
	Payload []byte
```

merlimat: The array can already be `nil`, no need for extra pointer

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+	Stat    Stat
+}
+
+type OxiaClient interface {
+	io.Closer
+	Put(key string, payload *[]byte, expectedVersion int64) (Stat, error)
```
merlimat:
```suggestion
	Put(key string, payload []byte, expectedVersion *int64) (Stat, error)
```

client/memory_impl.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func (c inMemoryOxiaClient) Close() error {
+	return nil
+}
+
+func (c inMemoryOxiaClient) Put(key string, payload *[]byte, expectedVersion int64) (Stat, error) {
+	now := c.clock.NowMillis()
+	if value, ok := c.data[key]; ok {
```
merlimat: Instead of duplicating all this logic, we could reuse the `StandaloneRpcServer` from https://github.com/streamnative/oxia/pull/24/files#diff-3c24a78e510b73776fb5313f6093d82fc05af4665e6d369edd5929c312c67967R16 , and passing the in-memory KVFactory.

client/memory_impl_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+func assertEquals[T any](t *testing.T, actual T, expected T) {
+	if !reflect.DeepEqual(actual, expected) {
+		t.Errorf("Stat did not match. Got %+v, expected %+v", actual, expected)
+	}
+}
+
+func assertError(t *testing.T, actual error, expected error) {
```
merlimat: Use test util methods from `testify/assert` package, as in https://github.com/streamnative/oxia/blob/main/server/kv/db_test.go#L67

client/client.go
```diff
@@ -0,0 +1,44 @@
...
+	CreatedTimestamp  uint64
+	ModifiedTimestamp uint64
+}
+
+type Value struct {
+	Payload *[]byte
+	Stat    Stat
+}
+
+type OxiaClient interface {
```
merlimat: I'd actually name this interface `Client` and living in the `oxia` package. 

Users would be calling this as 
```go
client, err := oxia.NewClient(&oxia.ClientOptions{
    ServiceURL: "localhost:9190",
})
```

client/client.go
merlimat: taking a pointer to allow passing `nil` as for "no expected version"

client/client.go
nahguam: good point 👍 , moved.

client/client.go
nahguam: That was the Java in me wanting to protect the var. 😬

client/memory_impl.go
nahguam: Yes, this was more for my learning, I can follow up with that later

client/client.go
nahguam: ah nice

oxia/memory_client.go
```diff
@@ -0,0 +1,89 @@
...
+		data:  make(map[string]Value),
+	}
+}
+
+func (c memoryClient) Close() error {
+	return nil
+}
+
+func (c memoryClient) Put(key string, payload []byte, expectedVersion *int64) (Stat, error) {
+	now := c.clock.NowMillis()
```
andrasbeni, line 26: Would it be possible to use the `NewPebbleKVFactory(testKVOptions)` to implement the in-memory key-value store? 
I know it would add some dependencies that should not exist in the final product, but I guess this `memoryClient` is only for testing purposes.

nahguam: Yes, @merlimat already suggested that. It'll get done in a follow up. I wanted to limit the scope while I'm learning go.

oxia/memory_client.go
```diff
@@ -0,0 +1,89 @@
+package oxia
```
merlimat, line 1: We should rename the  with `_test.go` so that is not included in production code.

oxia/memory_client.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func newMemoryClientWithClock(clock common.Clock) Client {
+	return &memoryClient{
+		clock: clock,
+		data:  make(map[string]Value),
+	}
+}
+
+func (c memoryClient) Close() error {
```
merlimat:
```suggestion
func (c *memoryClient) Close() error {
```

In general, we should almost always use the pointer dispatcher instead of the “value”, to avoid making a copy of the struct each time the method is called. Also, there is no effect on the original object since it operates just on the copy. 

I know, it’s a bit weird 🙈

nahguam: yea, still getting my head around this one.

oxia/memory_client.go
nahguam: can you clarify?

merlimat: Sorry, I was even missing words in there :)

just that everything test related should be in `xxx_test.go` files.

in any case, not important here because memoryClient is WIP for the client code

nahguam: do you mean changing the package to `oxia_test` in the test `*_test.go` files?

nahguam: That will also mean make some things public that we don't need to be. I see some blogs talking about having stuff in an internal package so that it's obvious to users that it's not for public consumption.

PR #27, opened 2022-11-10
Use shard-id field in all internal RPC requests

merlimat:
Passing the shard id in the GRPC metadata is not ideal. In general all the RPC requests will take the `shardId` field. The only request that cannot take it is the `AddEntries` since it sends a stream

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .idea
-oxia/cmd
-oxia-client
+bin/oxia
 **/*.pb.go
 .DS_Store
 **/*.toolbox
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,6 @@ FROM alpine:3.16.1
 RUN mkdir /oxia
 WORKDIR /oxia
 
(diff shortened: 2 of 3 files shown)


server/shard_manager.go
```diff
@@ -617,24 +613,25 @@ func (s *shardManager) addEntryResponseSync(follower string, response *proto.Add
...
+			} else {
+				srv.Send(resp)
+				return nil, nil
+			}
 		})
-		err = srv.Send(response)
-		if err != nil {
-			return nil, err
-		}
+		return srv.Send(response)
```
andrasbeni, line 629: I think this is unnecessary because we send the response from within the command.
Additionally, the event loop will not respond back through the command's reponseChan if the response is `nil`, so we would never get to execute this line of code. 
This means that `enqueueCommandAndWaitForResponseWithChannel` blocks indefinitely. Hmm. Something wasn't thought through here...

merlimat: Good catch. fixed it. 

I think we should try to not use channel in the beginning, unless for things where we do want to have buffering.

PR #29, opened 2022-11-10
Client side shard management

nahguam:


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/rs/zerolog v1.26.0
 	github.com/spf13/cobra v1.3.0
 	github.com/stretchr/testify v1.7.0
+	github.com/zeebo/xxh3 v1.0.2
```
merlimat, line 11: We have already https://github.com/cespare/xxhash in the dependencies list. Perhaps we can use that (although I don't known which one is better/faster)

common/hash.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+type Hash interface {
+	Hash(key string) uint32
+}
+
+type xxhash6432 struct {
+}
+
+func NewXXHash6432() Hash {
```
merlimat: Instead of using interface and struct, we could just pass a `func (string) uint32`

oxia/shard_manager.go
```diff
@@ -0,0 +1,150 @@
...
+	"oxia/common"
+	"oxia/proto"
+	"sync"
+)
+
+var (
+	ErrorShardNotFound     = errors.New("Shard not found")
+	ErrorUnknownShardRange = errors.New("Unknown toShard range")
+
+	logger = log.With().
```
merlimat: variables are visible inside the package, so we should have a `log` field in the `shardManager` struct.

go.mod
nahguam: `cespare/xxhash` is `XXH64`. `zeebo/xxh3` is `XXH3` which is the new variant of xxHash and is supposed to be faster
http://cyan4973.github.io/xxHash/

merlimat: 👍 Gotcha. Then (later) we should rename the enum in https://github.com/streamnative/oxia/blob/main/proto/client.proto#L98

oxia/client.go
```diff
@@ -10,10 +10,21 @@ const (
...
-	ErrorBadVersion  = errors.New("Bad version")
+	ErrorKeyNotFound   = errors.New("key not found")
+	ErrorBadVersion    = errors.New("bad version")
+	ErrorUnknownStatus = errors.New("unknown status")
+	ErrorShuttingDown  = errors.New("shutting down")
 )
 
+type Client interface {
+	io.Closer
+	Put(key string, payload []byte, expectedVersion *int64) <-chan PutResult
```
merlimat, line 21: Since these are all wrappers around a `(result, err)`, I'd just use a tuple here instead of defining types. 

eg: 

```go
Put(key string, payload []byte, expectedVersion *int64) <-chan (Stat, error)
```

oxia/convert.go
```diff
@@ -0,0 +1,35 @@
+package oxia
```
merlimat: Move to `oxia/internal` package

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
+package oxia
```
merlimat: Move to `oxia/internal`

oxia/convert.go
```diff
@@ -0,0 +1,35 @@
...
+	} else {
+		return Shard{
+			Id:        assignment.ShardId,
+			Leader:    assignment.Leader,
+			HashRange: hashRange,
+		}, nil
+	}
+}
+
+// TODO handle different shard range types
```
merlimat: Since it's the only scheme we support so far, we can remove the TODO
```suggestion
```

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+	"context"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"oxia/common"
+	"oxia/proto"
+	"sync"
+)
+
+type shardManager interface {
```
merlimat:
```suggestion
type ShardManager interface {
```

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+	start() <-chan bool
```
merlimat: Use capitalized method names in interfaces (as in: they're "public)

```suggestion
	Start() <-chan bool
```

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+	defer s.Unlock()
+
+	predicate := s.shardStrategy.Get(key)
+
+	for _, shard := range s.shards {
+		if predicate(shard) {
+			return shard.Id, nil
+		}
+	}
+	return 0, ErrorShardNotFound
```
merlimat: This in theory should be "impossible" :), if we validate the shards split when we receive them

oxia/client.go
merlimat, line 21: I really like the APIs returning the channel! It makes it very clear and simple for the user to choose how to use it

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+	shards        map[uint32]Shard
+	logger        zerolog.Logger
+}
+
+func newShardManager(shardStrategy ShardStrategy, clientPool common.ClientPool, serviceUrl string) shardManager {
+	return &shardManagerImpl{
+		shardStrategy: shardStrategy,
+		clientPool:    clientPool,
+		serviceUrl:    serviceUrl,
+		ready:         make(chan bool, 1),
```
merlimat: For this purpose an "unbuffered" channel should be good enough: `make(chan bool)`

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+		shards:        make(map[uint32]Shard),
+		logger:        log.With().Str("component", "shardManager").Logger(),
+	}
+}
+
+func (s *shardManagerImpl) start() <-chan bool {
+	go s.receiveWithRecovery()
+	return s.ready
```
merlimat: nit: instead of using `s.ready` as a member variable, we could create a new channel here: 

```go
readyCh := make(chan bool)
go s.receiveWithRecovery(readyCh)
return readyCh
```

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/rs/zerolog/log"
+	"oxia/common"
+	"oxia/proto"
+	"sync"
+)
+
+type shardManager interface {
+	start() <-chan bool
+	get(key string) (uint32, error)
+	getAll() []uint32
```
merlimat: What would be the usage of returning all the shard ids?

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+		return shard.Leader, nil
+	} else {
+		return "", ErrorShardNotFound
+	}
+}
+
+func (s *shardManagerImpl) receiveWithRecovery() {
+	for {
+		if err := <-s.receive(); err != nil {
+			//TODO fail on non-retryable errors
```
merlimat: There should be a way to tear this go routine down

oxia/shard_manager.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+	return 0, ErrorShardNotFound
+}
+
+func (s *shardManagerImpl) getAll() []uint32 {
+	s.Lock()
+	defer s.Unlock()
+
+	shardIds := make([]uint32, 0)
```
andrasbeni:
```suggestion
	shardIds := make([]uint32, len(s.shards))
	
```

go.mod
nahguam: done it now

oxia/client.go
nahguam: I don't think that's possible. It was my initial plan but `chan (...)` isn't allowed. My understanding is that `(...)` isn't actually a tuple, it's just the way to express returning multiple values from a func.

oxia/shard_manager.go
nahguam: panic instead?

oxia/convert.go
nahguam: I agree with the motivation here, but I'm having trouble achieving this in practice.

If I move non-public API things to package `oxia/oxia/internal`, they still need to operate on things in the public API, mostly the structs, so the internal package needs to `import "oxia/oxia"`. However, the components in the public package need to consume the utilities, so the public package needs to import `import "oxia/oxia/internal"` - hence import cycle. 🤷 

This is why I had kept things in the oxia package but not exposed them publicly, ie. lowercase

nahguam: I think I'm going to have to move the common imported types/funcs to a third package to break the cycle

nahguam: ok, got this sorted 👍

oxia/shard_manager.go
nahguam: `put`, `delete` and `get` all use the key to identify the shard (and leader) to send the request to.

`delete_range` and `get_range` need to send a request to all shards.

oxia/convert.go
nahguam: What I've done is kept the public interfaces and structs in `oxia`, shard manager and convert in `oxia/internal` and the publicly accessible implementations in `oxia/client`. I feel this is the best compromise.

It avoids the import cycle but it does mean that users creating a new client will do:

```
import "oxia/oxia/client"
...

options := client.NewClientOptions()
...
```

i.e. they won't be able to do `oxia.NewClientOptions()` unless they alias their import.

The other option would be to put all the public interfaces in `oxia/api` and leave the impl in `oxia`. Thoughts?

nodece: Why not use the `client` package?

nahguam: > Why not use the `client` package? 

For a 3rd party bringing this in as a dependency, `client` is too generic.

I'm now thinking the public API should go in in `oxia/api` and the impls in `oxia` to keep with that original thinking.

nodece: I suggest we make an `internal` package in the root, and move the `client`, `server`, and `standalone` to here.

PR #30, opened 2022-11-14
Added refactored code for follower controller and tests

merlimat:
Moved the logic for the "follower" into its own separate controller. 

Changes:
 * Separate follower from leader logic, to simplify the controller
 * Use mutex instead of channel to make code easier. We should use channel only where we want to have some buffering
 * Removed go-routine from follower logic (although we should probably add one to scan the wal in background, instead of doing it before acking back to the leader).
 * Added initial set of unit tests for the  follower

### DIFF
--- a/server/follower_controller_test.go
+++ b/server/follower_controller_test.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"fmt"
 	"github.com/stretchr/testify/assert"
 	pb "google.golang.org/protobuf/proto"
 	"oxia/proto"
@@ -17,7 +16,7 @@ var testKVOptions = &kv.KVFactoryOptions{
 func TestFollower(t *testing.T) {
 	var shardId uint32
 	kvFactory := kv.NewPebbleKVFactory(testKVOptions)
-	wal := NewInMemoryWal(ShardId(fmt.Sprint(shardId)))
+	wal := NewInMemoryWal(shardId)
 
 	fc, err := NewFollowerController(shardId, wal, kvFactory)
 	assert.NoError(t, err)
@@ -100,7 +99,7 @@ func TestFollower(t *testing.T) {
 	assert.EqualValues(t, 3, fc.Epoch())
 
 	// Double-check the values in the DB
-	dbRes, err := fc.(*followerController).db.ProcessBatch(&proto.BatchRequest{Gets: []*proto.GetRequest{{
+	dbRes, err := fc.(*followerController).db.ProcessRead(&proto.ReadRequest{Gets: []*proto.GetRequest{{
 		Key:            "a",
 		IncludePayload: true}, {
 		Key:            "b",
@@ -120,7 +119,7 @@ func TestFollower(t *testing.T) {
 func TestReadingUpToCommitIndex(t *testing.T) {
 	var shardId uint32
 	kvFactory := kv.NewPebbleKVFactory(testKVOptions)
-	wal := NewInMemoryWal(ShardId(fmt.Sprint(shardId)))
+	wal := NewInMemoryWal(shardId)
 
 	fc, err := NewFollowerController(shardId, wal, kvFactory)
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


server/follower_controller.go
```diff
@@ -0,0 +1,261 @@
...
+	return fc.status
+}
+
+func (fc *followerController) Epoch() uint64 {
+	fc.Lock()
+	defer fc.Unlock()
+	return fc.epoch
+}
+
+func (fc *followerController) Fence(req *proto.FenceRequest) (*proto.FenceResponse, error) {
```
andrasbeni, line 126: Did you omit synchronization intentionally here?

server/follower_controller.go
```diff
@@ -0,0 +1,261 @@
...
+			return err
+		}
+	}
+}
+
+func (fc *followerController) addEntry(req *proto.AddEntryRequest) (*proto.AddEntryResponse, error) {
+	fc.Lock()
+	defer fc.Unlock()
+
+	if fc.status != Follower && fc.status != Fenced {
```
andrasbeni, line 183: Does this check make any sense now? I think `status == Follower || status == Fenced` is an invariant of `followerController`, isn't it. 
What do you think about reducing the status to a `bool` flag `isFenced`? In my opinion, it would make the code more readable but would be harder to relate the code to the TLA specification.

server/follower_controller_test.go
```diff
@@ -0,0 +1,293 @@
...
+	response = stream.GetResponse()
+	assert.EqualValues(t, 0, response.Epoch)
+	assert.Nil(t, response.EntryId)
+	assert.True(t, response.InvalidEpoch)
+
+	assert.Equal(t, Follower, fc.Status())
+	assert.EqualValues(t, 1, fc.Epoch())
+
+	// Try to add entry with higher epoch
+	stream.AddRequest(createAddRequest(t, 2, 0, map[string]string{"a": "4", "b": "5"}, &proto.EntryId{
```
andrasbeni: According to [the spec](https://github.com/streamnative/oxia/blob/main/tlaplus/OxiaReplication.tla#L1190), higher than known epoch is not a case for `InvalidEpoch`.

server/follower_controller_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Follower needs to be in "Fenced" state to receive a Truncate request
+	tr, err := fc.Truncate(&proto.TruncateRequest{
+		Epoch: 1,
+		HeadIndex: &proto.EntryId{
+			Epoch:  0,
+			Offset: 0,
+		},
+	})
+	assert.ErrorIs(t, err, ErrorInvalidStatus)
```
andrasbeni: Not sure about go practices, but it seems to me that in this test not only "reading up to commit index" is tested. Can you please extract the `InvalidStatus` part into a separate test function?

server/follower_controller_test.go
```diff
@@ -0,0 +1,293 @@
...
+	assert.EqualValues(t, 1, fc.Epoch())
+
+	tr, err := fc.Truncate(&proto.TruncateRequest{
+		Epoch: 2,
+		HeadIndex: &proto.EntryId{
+			Epoch:  0,
+			Offset: 0,
+		},
+	})
+	assert.Nil(t, tr)
```
andrasbeni, line 234: Didn't you mean to assert that the epoch did not change? At leat that's what the test function's name suggests

server/follower_controller.go
merlimat: No, here I forgot :)

server/follower_controller.go
merlimat: Let's leave it with the codes for now, so that we can differentiate the `NotMember` state.

server/follower_controller_test.go
merlimat: Truncate with the wrong epoch will be ignored and the follower stays in `Fenced` state

server/follower_controller_test.go
merlimat: Fixed

server/follower_controller_test.go
merlimat: Moved the individual case out

PR #33, opened 2022-11-15
Split client batch to separate writes and reads

nahguam:


proto/client.proto
```diff
@@ -115,28 +123,46 @@ message BatchRequest {
...
   // The put responses
   repeated PutResponse puts = 1;
   // The delete responses
   repeated DeleteResponse deletes = 2;
   // The delete range responses
   repeated DeleteRangeResponse delete_ranges = 3;
+}
+
+/**
+ * A batch read request. Applies the batches of requests. Requests are processed
```
andrasbeni: Does it matter what the order of processing is?

proto/client.proto
```diff
@@ -207,9 +233,9 @@ message GetResponse {
 }
 
 /**
- * Input to a delete range or get range request.
+ * Input to a get range request.
  */
-message RangeRequest {
+message DeleteRangeRequest {
```
andrasbeni, line 236: I like the separation of DeleteRangeReq and GetRangeReq, but it wasn't strictly necessary, was it.

server/kv/db.go
```diff
@@ -67,30 +68,38 @@ func (d *db) ProcessBatch(b *proto.BatchRequest) (*proto.BatchResponse, error) {
...
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func (d *db) ProcessRead(b *proto.ReadRequest) (*proto.ReadResponse, error) {
+	res := &proto.ReadResponse{}
+
+	kv := d.kv
```
andrasbeni, line 86: _If_ we want to guarantee read consistency (e.g. a range scan and a get return the same value for the same key), we will need some kind of `d.kv.NewReadBatch()`. It would probably use a pebble iterator in the end.

proto/client.proto
nahguam: not for reads, no. I'll remove

proto/client.proto
nahguam: no but helps being able to switch with a type assertion on the request type.

server/kv/db.go
nahguam: I've added a TODO

PR #34, opened 2022-11-15
Add golangci-linter

nodece:
Changes:

- Add golangci-lint to GitHub Actions
- Fix lint error

### DIFF
--- a/common/signal.go
+++ b/common/signal.go
@@ -13,20 +13,18 @@ func WaitUntilSignal(closer io.Closer) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
-	for {
-		sig := <-c
-		log.Info().
-			Str("signal", sig.String()).
-			Msg("Received signal, exiting")
-		err := closer.Close()
-		if err != nil {
-			log.Error().
-				Err(err).
-				Msg("Failed when shutting down server")
-			os.Exit(1)
-		} else {
-			log.Info().Msg("Shutdown Completed")
-			os.Exit(0)
-		}
+	sig := <-c
+	log.Info().
+		Str("signal", sig.String()).
+		Msg("Received signal, exiting")
+	err := closer.Close()
+	if err != nil {
+		log.Error().
+			Err(err).
+			Msg("Failed when shutting down server")
+		os.Exit(1)
+	} else {
+		log.Info().Msg("Shutdown Completed")
+		os.Exit(0)
 	}
 }
--- a/server/kv/db_test.go
+++ b/server/kv/db_test.go
@@ -98,8 +98,8 @@ func TestDBSimple(t *testing.T) {
 	r8 := res.Deletes[3]
 	assert.Equal(t, proto.Status_KEY_NOT_FOUND, r8.Status)
 
+	// TODO: Add the request call and the verification
 	/// Second batch
-
 	//req = &proto.WriteRequest{
 	//	Puts: []*proto.PutRequest{
 	//		{ // Should succeed: no version check
--- a/server/utils_test.go
+++ b/server/utils_test.go
@@ -18,12 +18,12 @@ func failOnErr(t *testing.T, err error, op string) {
 	}
 }
 
-func newWalWithEntries(payloads ...string) *inMemoryWal {
+func newWalWithEntries(payloads ...string) (*inMemoryWal, error) {
 	wal := NewInMemoryWal(shard)
 	return initWalWithEntries(wal, payloads)
 }
 
-func initWalWithEntries(wal Wal, payloads []string) *inMemoryWal {
+func initWalWithEntries(wal Wal, payloads []string) (*inMemoryWal, error) {
 	epoch := uint64(1)
 	offset := uint64(1)
 	for _, p := range payloads {
@@ -40,10 +40,10 @@ func initWalWithEntries(wal Wal, payloads []string) *inMemoryWal {
 			}
 			err := wal.Append(entry)
 			if err != nil {
-				panic(err)
+				return nil, err
 			}
 			offset++
 		}
 	}
-	return wal.(*inMemoryWal)
+	return wal.(*inMemoryWal), nil
 }
--- a/server/shard_manager_test.go
+++ b/server/shard_manager_test.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"testing"
+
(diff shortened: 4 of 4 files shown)


common/signal.go
```diff
@@ -1,20 +1,20 @@
...
 	"os/signal"
 	"syscall"
-	"time"
+
+	"github.com/rs/zerolog/log"
 )
 
 func WaitUntilSignal(closer io.Closer) {
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
merlimat, line 13: The intention here was to declare an unbuffered channel

cmd/main.go
```diff
@@ -13,9 +13,9 @@ import (
 
 var (
 	// Used for flags.
-	shards            uint32
-	staticNodes       []string
-	replicationFactor uint32
+	//shards            uint32
+	//staticNodes       []string
+	//replicationFactor uint32
```
merlimat: We should removed all the vars instead of commenting it out

server/kv/db_test.go
```diff
@@ -100,45 +100,45 @@ func TestDBSimple(t *testing.T) {
...
-				Key:             "c",
-				ExpectedVersion: pb.Int64(-1),
-			},
-			{ // Should fail: the key was not inserted before the batch
-				Key:             "c",
-				ExpectedVersion: pb.Int64(0),
-			},
-		},
-	}
+	//req = &proto.WriteRequest{
```
merlimat, line 103: We shouldn't comment out this test

server/kv/kv_pebble_test.go
```diff
@@ -17,9 +17,9 @@ func TestPebbbleSimple(t *testing.T) {
 	assert.NoError(t, err)
 
 	wb := kv.NewWriteBatch()
-	wb.Put("a", []byte("0"))
-	wb.Put("b", []byte("1"))
-	wb.Put("c", []byte("2"))
+	_ = wb.Put("a", []byte("0"))
```
merlimat: It would be better to convert these calls into: 

```suggestion
	assert.NoError(t, wb.Put("a", []byte("0")))
```

server/main.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 var (
-	conf        = &serverConfig{}
-	staticNodes []string
+	conf = &serverConfig{}
+	//staticNodes []string
```
merlimat:
```suggestion
```

server/server.go
```diff
@@ -42,7 +42,7 @@ type server struct {
 	shardsDirector ShardsDirector
 	clientPool     common.ClientPool
 
-	identityInternalAddress proto.ServerAddress
+	//identityInternalAddress proto.ServerAddress
```
merlimat:
```suggestion
```

server/utils_test.go
```diff
@@ -38,7 +38,7 @@ func initWalWithEntries(wal Wal, payloads []string) *inMemoryWal {
 				Value:     []byte(p),
 				Timestamp: offset,
 			}
-			wal.Append(entry)
+			_ = wal.Append(entry)
```
merlimat:
```suggestion
			assert.NoError(t, wal.Append(entry))
```

standalone/standalone.go
```diff
@@ -19,7 +18,7 @@ type standalone struct {
 	rpc       *StandaloneRpcServer
 	kvFactory kv.KVFactory
 
-	identityInternalAddress proto.ServerAddress
+	//identityInternalAddress proto.ServerAddress
```
merlimat:
```suggestion
```

common/signal.go
nodece: We need to fix the `sigchanyzer: misuse of unbuffered os.Signal channel as argument to signal.Notify (govet)`.

server/kv/db_test.go
nodece: This `req` unused.

server/utils_test.go
nodece: Use `panic` instead.

server/shard_manager.go
```diff
@@ -626,7 +626,10 @@ func (s *shardManager) AddEntries(srv proto.OxiaLogReplication_AddEntriesServer)
 				return resp, nil
 			}
 		})
-		return srv.Send(response)
+		err = srv.Send(response)
```
nodece, line 629: Please confirm this code.

server/utils_test.go
merlimat: This is for tests, we use `assert` module to check all expected results

common/signal.go
merlimat: From that discussion in https://github.com/golang/go/issues/45604, it looks to be a false positive, right?

merlimat: Anyway, it should work fine in both ways

server/kv/db_test.go
merlimat: Ouch.. then leave it commented and please add a `// TODO:` comment so that we don't forget to add the request call and the verification

server/shard_manager.go
merlimat: ~I don't understand the difference between the two of them :-)~

server/utils_test.go
```diff
@@ -38,7 +38,10 @@ func initWalWithEntries(wal Wal, payloads []string) *inMemoryWal {
 				Value:     []byte(p),
 				Timestamp: offset,
 			}
-			wal.Append(entry)
+			err := wal.Append(entry)
```
merlimat, line 41: Why not using `asserts.NoError(t, wal.Append(entry))`?

merlimat: Ok, the method itself should be taking a `t *testing.T` argument

server/utils_test.go
merlimat: Ok.. I kind of see your point now.. `panic` is ok to use here

common/signal.go
nodece: Yes, it works fine.

server/shard_manager.go
merlimat: Yes, it's the correct change here

server/utils_test.go
nodece: Use the `panic` or `asserts.NoError` is fine with me.

nodece: Let me return this error.

PR #36, opened 2022-11-15
WAL API change

andrasbeni:


server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
 	"oxia/proto"
 )
 
+var (
+	ErrorEntryNotFound = errors.New("oxia: entry not found")
+	ErrorReaderClosed  = errors.New("oxia: reader already closed")
+)
+
+type Reader interface {
```
merlimat: Maybe `WalReader`?

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
+	io.Closer
+	ReadNext() (*proto.LogEntry, error)
```
merlimat: We should also have `HasNext() bool`, which we'd be using when replaying the WAL on startup.

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
+}
+
 type Wal interface {
 	io.Closer
 	Append(entry *proto.LogEntry) error
-
-	Read(lastPushedEntryId EntryId, callback func(*proto.LogEntry) error) error
-	ReadSync(previousCommittedEntryId EntryId, lastCommittedEntryId EntryId, callback func(*proto.LogEntry) error) error
 	GetHighestEntryOfEpoch(epoch uint64) (EntryId, error)
```
merlimat: I'd keep this out of the WAL logic. Eg: we could create a reverse iterator and do it on top of it.

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
 	LogLength() uint64
-	EntryIdAt(index uint64) EntryId
 }
 
 type inMemoryWal struct {
-	shard     uint32
-	log       []*proto.LogEntry
-	index     map[EntryId]int
-	callbacks []func(*proto.LogEntry) error
+	shard        uint32
```
merlimat: I'd move the `inMemoryWal` to its own file, perhaps scope for tests only `wal_memory_test.go`

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
+	ErrorEntryNotFound = errors.New("oxia: entry not found")
+	ErrorReaderClosed  = errors.New("oxia: reader already closed")
+)
+
+type Reader interface {
+	io.Closer
+	ReadNext() (*proto.LogEntry, error)
+}
+
 type Wal interface {
```
merlimat: We could have a `WalFactory` (like https://github.com/streamnative/oxia/blob/main/server/kv/kv.go#L65) so that we can have different kind of factories that are passed in, and each is able to create the Wal for a given shard.

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
 	io.Closer
 	Append(entry *proto.LogEntry) error
-
-	Read(lastPushedEntryId EntryId, callback func(*proto.LogEntry) error) error
-	ReadSync(previousCommittedEntryId EntryId, lastCommittedEntryId EntryId, callback func(*proto.LogEntry) error) error
 	GetHighestEntryOfEpoch(epoch uint64) (EntryId, error)
 	TruncateLog(headIndex EntryId) (EntryId, error)
-	StopReaders()
-	ReadOne(id EntryId) (*proto.LogEntry, error)
+	CloseReaders() error
```
merlimat: I think we shouldn't be needing to manage the readers from here. 

Readers should be closed independently, before the Wal is closed().

server/wal.go
```diff
@@ -7,97 +7,108 @@ import (
...
+	Reader(firstOffset uint64) (Reader, error)
```
merlimat: Maybe 
```suggestion
	NewReader(startOffset uint64) (Reader, error)
```

server/follower_controller.go
```diff
@@ -213,29 +215,79 @@ func (fc *followerController) addEntry(req *proto.AddEntryRequest) (*proto.AddEn
...
 
-		_, err := fc.db.ProcessWrite(br)
-		return err
-	})
+	return nil, nil
+
+}
 
+func (fc *followerController) getHighestEntryOfEpoch(epoch uint64) (wal.EntryId, error) {
+	// TODO duplicated between here and shard_manager.go
```
merlimat: This could be a simple function and reused across the 2: 

```go
func getHighestEntryOfEpoch(wal Wal, epoch uint64) (wal.EntryId, error) {
}
```

server/wal/wal.go
```diff
@@ -0,0 +1,60 @@
...
+type WalReader interface {
+	io.Closer
+	// ReadNext returns the next entry in the log according to the Reader's direction.
+	// If a reverse WalReader has passed the beginning of the log, it returns [ErrorEntryNotFound]. To avoid this error, use HasNext.
+	// If a forward WalReader has passed the log end, it will wait for new entries appended. To avoid waiting, you can use HasNext.
+	ReadNext() (*proto.LogEntry, error)
+	// HasNext returns true if there is an entry to read.
+	// For a reverse WalReader this means the reader has not yet reached the beginning of the log.
+	// For a forward WalReader this means that we have not yet reached the offset that was the log end when the reader was created.
+	HasNext() (bool, error)
```
merlimat: Returning `error` here complicates how we iterate over the wal. I'd say to just return `bool`. 

Any implementation should be able to read the min/max offsets when it's started and then be able to answer `HasNext()` without doing any IO operation.

server/follower_controller.go
```diff
@@ -238,6 +233,69 @@ func (fc *followerController) addEntry(req *proto.AddEntryRequest) (*proto.AddEn
 
 }
 
+func (fc *followerController) processCommittedEntries(offset uint64) error {
+	fmt.Printf("ProcessCommitted (%d)", offset)
```
merlimat:
```suggestion
	fc.log.Debug().
	     UInt64("offset", offset).
	     Msg("ProcessCommitted)
```

server/follower_controller.go
```diff
@@ -238,6 +233,69 @@ func (fc *followerController) addEntry(req *proto.AddEntryRequest) (*proto.AddEn
...
+	}
+	defer func() {
+		err := reader.Close()
+		if err != nil {
+			fc.log.Err(err).Msg("Error closing reader used for applying committed entries")
+		}
+	}()
+
+	for limit := fc.commitIndex.Offset; offset <= limit && reader.HasNext(); offset++ {
+		fmt.Printf("Reading offset (%d)", offset)
```
merlimat: Debug or remove

andrasbeni: Removed. Sorry, I tend to litter the code when desperately debugging and I knew it is not going to be merged.

PR #40, opened 2022-11-17
Client implementation

nahguam:
This is stacked on https://github.com/streamnative/oxia/pull/39

### DIFF
--- a/oxia/async_client_impl_test.go
+++ b/oxia/async_client_impl_test.go
@@ -24,45 +24,45 @@ func TestAsyncClientImpl(t *testing.T) {
 		BatchMaxSize: 1,
 		BatchTimeout: DefaultBatchTimeout,
 	}
-	oxiaClient := NewAsyncClient(options)
+	client := NewAsyncClient(options)
 
-	putResult := <-oxiaClient.Put("/a", []byte{0}, &VersionNotExists)
+	putResult := <-client.Put("/a", []byte{0}, &VersionNotExists)
 	assert.Equal(t, versionZero, putResult.Stat.Version)
 
-	getResult := <-oxiaClient.Get("/a")
+	getResult := <-client.Get("/a")
 	assert.Equal(t, GetResult{
 		Payload: []byte{0},
 		Stat:    putResult.Stat,
 	}, getResult)
 
-	putResult = <-oxiaClient.Put("/c", []byte{0}, &VersionNotExists)
+	putResult = <-client.Put("/c", []byte{0}, &VersionNotExists)
 	assert.Equal(t, versionZero, putResult.Stat.Version)
 
-	putResult = <-oxiaClient.Put("/c", []byte{1}, &versionZero)
+	putResult = <-client.Put("/c", []byte{1}, &versionZero)
 	assert.Equal(t, int64(1), putResult.Stat.Version)
 
-	getRangeResult := <-oxiaClient.GetRange("/a", "/d")
+	getRangeResult := <-client.GetRange("/a", "/d")
 	assert.Equal(t, GetRangeResult{
 		Keys: []string{"/a", "/c"},
 	}, getRangeResult)
 
-	deleteErr := <-oxiaClient.Delete("/a", &versionZero)
+	deleteErr := <-client.Delete("/a", &versionZero)
 	assert.ErrorIs(t, nil, deleteErr)
 
-	getResult = <-oxiaClient.Get("/a")
+	getResult = <-client.Get("/a")
 	assert.Equal(t, GetResult{
 		Err: ErrorKeyNotFound,
 	}, getResult)
 
-	deleteRangeResult := <-oxiaClient.DeleteRange("/c", "/d")
+	deleteRangeResult := <-client.DeleteRange("/c", "/d")
 	assert.ErrorIs(t, nil, deleteRangeResult)
 
-	getResult = <-oxiaClient.Get("/d")
+	getResult = <-client.Get("/d")
 	assert.Equal(t, GetResult{
 		Err: ErrorKeyNotFound,
 	}, getResult)
 
-	err = oxiaClient.Close()
+	err = client.Close()
 	assert.ErrorIs(t, nil, err)
 
 	err = server.Close()


common/client_pool.go
```diff
@@ -80,19 +81,23 @@ func (cp *clientPool) getConnection(target string) (grpc.ClientConnInterface, er
...
+		return cnx.(grpc.ClientConnInterface), nil
+	}
+
 	cp.log.Info().
 		Str("server_address", target).
 		Msg("Creating new GRPC connection")
 
 	cnx, err := grpc.Dial(target, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		existing, loaded := cp.connections.LoadOrStore(target, cnx)
```
nahguam, line 89: This was bugged and creating a new connection on every call, so I added a double-check-lock

standalone/standalone_rpc_server.go
```diff
@@ -29,8 +30,9 @@ type StandaloneRpcServer struct {
 func NewStandaloneRpcServer(port int, identityAddr string, numShards uint32, kvFactory kv.KVFactory) (*StandaloneRpcServer, error) {
 	// Assuming 1 single shard
 	res := &StandaloneRpcServer{
-		numShards: numShards,
-		dbs:       make(map[uint32]kv.DB),
+		identityAddr: identityAddr + ":" + strconv.FormatInt(int64(port), 10),
```
nahguam: This wasn't wired in.

internal/client/test_utils.go
```diff
@@ -0,0 +1,21 @@
+package client
+
+import "net"
+
+func GetFreePort() int {
```
merlimat: This still has a race condition where multiple processes/tests in parallel end up trying to use the same port. 

A better solution would be to make the standalone server to be configured with 0 and reporting the actual port it ends up using.

common/client_pool.go
merlimat: Wouldn't the `Load()` at the beginning of method find it on the next time?

common/client_pool.go
```diff
@@ -18,6 +18,7 @@ type ClientPool interface {
 }
 
 type clientPool struct {
+	sync.Mutex
```
merlimat, line 21: If we have additional mutex, then we can convert `sync.Map` into a `map[string]grpc.ClientConnInterface`

internal/client/batch/test_utils.go
```diff
@@ -0,0 +1,17 @@
+package batch
```
merlimat: Maybe rename to `utils_test.go` to make it part of "test classpath" only.

common/client_pool.go
merlimat: (In any case I think that putting the mutex upfront is better approach, especially if we can then cache the connection object for a given shard)

internal/client/batch/batch.go
```diff
@@ -0,0 +1,8 @@
+package batch
+
+type batch interface {
+	add(any)
```
merlimat: The methods of an interface are all meant the be "public" (within the context of visibility of the interface itself) and we should use capitalized names.

merlimat: Same for the interface name

internal/client/batch/batcher.go
```diff
@@ -0,0 +1,94 @@
...
+func (b *BatcherFactory) NewReadBatcher(shardId *uint32) Batcher {
+	return b.newBatcher(shardId, readBatchFactory{execute: b.Executor.ExecuteRead}.newBatch)
+}
+
+func (b *BatcherFactory) newBatcher(shardId *uint32, batchFactory func(shardId *uint32) batch) Batcher {
+	batcher := &batcherImpl{
+		shardId:      shardId,
+		batchFactory: batchFactory,
+		c:            make(chan any, b.MaxSize), //TODO maybe unbuffered?
+		close:        make(chan bool),
```
merlimat: A convention I've seen used for naming channel variables is the `ch` suffix.

In this case:
 * `c` -> `requestsCh`
 * `close` -> `closeCh`

internal/client/batch/manager.go
```diff
@@ -0,0 +1,46 @@
...
+
+type Manager struct {
+	sync.Mutex
+	batcherFactory func(*uint32) Batcher
+	batchers       map[uint32]Batcher
+}
+
+func (b *Manager) Get(shardId uint32) Batcher {
+	//double-check lock
+	batcher, ok := b.batchers[shardId]
```
merlimat: If the map gets indeed updated when shards change, we cannot read it without a mutex

nahguam: I think this map is thread safe, however you bring up a good point. There is no mechanism for existing open batches to be failed (or routed to the correct batcher for the new shard) and batchers to be cleaned up.

I think we can tackle that separately though.

oxia/client.go
```diff
@@ -3,20 +3,44 @@ package oxia
 import (
 	"errors"
 	"io"
+	"time"
 )
 
 const (
-	VersionNotExists int64 = -1
+	DefaultBatchLinger  = 5 * time.Millisecond
+	DefaultBatchMaxSize = 100
```
merlimat: nit: I'd go up to 1K

oxia/client.go
```diff
@@ -3,20 +3,44 @@ package oxia
...
+	DefaultBatchTimeout = 1 * time.Minute
```
merlimat: Maybe 30 seconds here?

oxia/client.go
```diff
@@ -3,20 +3,44 @@ package oxia
...
+type ClientOptions struct {
+	ServiceUrl   string
+	BatchLinger  time.Duration
+	BatchMaxSize int
+	BatchTimeout time.Duration
+	InMemory     bool
+}
+
+func NewAsyncClient(options *ClientOptions) AsyncClient {
+	if options.InMemory {
```
merlimat: Is this option temporary? Or when would a user select the in-memory client?

oxia/client.go
```diff
@@ -3,20 +3,44 @@ package oxia
...
+		return newMemoryClient()
+	}
+	return newAsyncClient(options)
+}
+
+func NewSyncClient(options *ClientOptions) SyncClient {
```
merlimat: (for subsequent PRs) 

Since we are taking the "serviceUrl", it doesn't really make sense to take `options = nil` which would mean "default options". 

Because of that, we could just take `options ClientOptions` as the argument. 

One problem with Go struct, when used as options, is that it's not easy to provide default values. All the integers default to 0, and pointers to `nil`. 


One aspect to consider is to that we might want to have `linger time = 0` when using sync API, or we're going to cap the throughput. 


There are 2 options, either we consider "0 means default" or we could

oxia/sync_client_impl.go
```diff
@@ -0,0 +1,40 @@
...
+		asyncClient: asyncClient,
+	}
+}
+
+func (c *syncClientImpl) Close() error {
+	return c.asyncClient.Close()
+}
+
+func (c *syncClientImpl) Put(key string, payload []byte, expectedVersion *int64) (Stat, error) {
+	r := <-c.asyncClient.Put(key, payload, expectedVersion)
```
merlimat, line 23: I wasn't 100% sure this was working fine if the channel was already closed, but it works perfectly. 

(The only problem is the panic when you're trying to *write* on a closed channel, reading is always fine)

internal/client/batch/manager.go
nahguam: Created #42 to track this

oxia/client.go
nahguam: removed this

oxia/client.go
nahguam: > or we could...

😄 

Yes, I was pondering this last night. I'll remove the pointer.

For the sync client, we can just do `max size = 1` then it'll send one at a time.

0 = default sounds like a reasonable starting point.

nahguam: Or I could try this - https://www.sohamkamani.com/golang/options-pattern/ or similar

I'll merge this PR do another one for that

oxia/async_client_impl_test.go
```diff
@@ -0,0 +1,70 @@
...
+	getResult = <-client.Get("/d")
+	assert.Equal(t, GetResult{
+		Err: ErrorKeyNotFound,
+	}, getResult)
+
+	err = client.Close()
+	assert.ErrorIs(t, nil, err)
+
+	err = server.Close()
+	assert.ErrorIs(t, nil, err)
```
andrasbeni, line 69: I know I'm a bit late. But I think this should be `assert.NoError(err)`. IIUC, this does something else

PR #41, opened 2022-11-18
Add commitIndex to DB write requests

andrasbeni:


server/kv/db.go
```diff
@@ -104,6 +124,30 @@ func (d *db) ProcessRead(b *proto.ReadRequest) (*proto.ReadResponse, error) {
...
+	getReq := &proto.GetRequest{
+		Key:            commitIndexKey,
+		IncludePayload: true,
+	}
+	gr, err := applyGet(kv, getReq)
+	if err != nil {
+		return nil, err
+	}
+	if gr.Status == proto.Status_KEY_NOT_FOUND {
+		return &proto.EntryId{}, nil
```
nahguam: `&proto.EntryId{}` - Would `nil` make more sense here?

andrasbeni: Right now a lot of code is implemented with the assumption that epoch 0 does not actually exist, so I use `EntryId{0,0}` to mean "nothing yet". I don't really like nil, so I try to avoid it. I think the next step will be to change epoch and offset from uint64 to uint32 and then I can use -1. But not this week.

common/keys.go
```diff
@@ -0,0 +1,4 @@
+package common
+
+// InternalKeyPrefix are reserved for use by oxia.
+const InternalKeyPrefix = "oxia/"
```
merlimat: We should probably add a `__` so that it's clearer it's an internal scope and it will also sort before the other keys

server/kv/db.go
merlimat: the `EntryId{0,0}` should be assigned to a constant like`NonExistentEntryId`

server/kv/db_test.go
```diff
@@ -377,3 +378,36 @@ func TestDBDeleteRange(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	commitIndex, err = db.ReadCommitIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, wal.EntryId{Epoch: epoch, Offset: offset},
+		wal.EntryIdFromProto(commitIndex))
+
+	assert.NoError(t, db.Close())
+	assert.NoError(t, factory.Close())
+}
```
merlimat, line 413: We should have a test with the snapshot. eg: 

 1. Write some entries `(e1, 2)` with `commitIndex1`
 2. Create a snapshot iterator
 3. Write more entries `(e3, e4)` with `commitIndex2`
 4. The snapshot should return `commitIndex1` and iterating over it should only return `(e1, e2)`

standalone/standalone_rpc_server.go
```diff
@@ -98,7 +98,7 @@ func (s *StandaloneRpcServer) ShardAssignments(_ *proto.ShardAssignmentsRequest,
 }
 
 func (s *StandaloneRpcServer) Write(ctx context.Context, write *proto.WriteRequest) (*proto.WriteResponse, error) {
-	return s.dbs[*write.ShardId].ProcessWrite(write)
+	return s.dbs[*write.ShardId].ProcessWrite(write, &proto.EntryId{})
```
merlimat: This could be dangerous as we can forget about it. There should be at least a TODO note.

server/kv/db.go
```diff
@@ -104,6 +124,30 @@ func (d *db) ProcessRead(b *proto.ReadRequest) (*proto.ReadResponse, error) {
 	return res, nil
 }
 
+func (d *db) ReadCommitIndex() (*proto.EntryId, error) {
+
+	// TODO ReadBatch?
```
merlimat: We don't need to read from a batch here

server/kv/db.go
```diff
@@ -79,6 +86,19 @@ func (d *db) ProcessWrite(b *proto.WriteRequest) (*proto.WriteResponse, error) {
 	return res, nil
 }
 
+func (d *db) addCommitIndex(commitIndex *proto.EntryId, batch WriteBatch) error {
+	commitIndexPayload, err := pb.Marshal(commitIndex)
+	if err != nil {
+		return err
+	}
+	_, err = applyPut(batch, &proto.PutRequest{
```
merlimat, line 95: There is a problem with this approach in that the `commitIndex` is only added to the DB but it's not added to the WAL storage entry. 

That would mean that after a crash, when we are replaying the WAL we would be be reconstructing the right state for the commit index.

The `PutRequest` should be added directly to the `WriteBatch` before inserting in the WAL.

server/kv/db.go
```diff
@@ -104,6 +124,30 @@ func (d *db) ProcessRead(b *proto.ReadRequest) (*proto.ReadResponse, error) {
 	return res, nil
 }
 
+func (d *db) ReadCommitIndex() (*proto.EntryId, error) {
```
merlimat, line 128: Since we need to add the entry in the `WriteBatch` and add it to the WAL, I'd just keep it completely out of the DB layer.

server/kv/db.go
andrasbeni: Sorry, I might be misunderstanding what you describe here.
The `commitIndex` we put here in the DB is the EntryId of the LogEntry we are applying. So after a crash ( if we still have the DB), we will not need to replay any entry before and including the `commitIndex`, only entries after it. 
And since the entry id is in the logEntry, it is redundant to add it as a Put inside the WriteBatch that is in that logEntry.

merlimat: You are absolutely right. I got all confused up 🥴

server/kv/db_test.go
andrasbeni: This is tested at the `KV` level in [kv_pebble_test.go](https://github.com/streamnative/oxia/blob/main/server/kv/kv_pebble_test.go#L115) and for the KV the commitIndex is just another key-value pair. At the `DB` level, taking a snapshot is not yet defined.

PR #44, opened 2022-11-18
Client options with defaults and validation

nahguam:


oxia/async_client_impl_test.go
```diff
@@ -18,12 +18,8 @@ func TestAsyncClientImpl(t *testing.T) {
...
 	assert.ErrorIs(t, nil, err)
 
-	options := ClientOptions{
-		ServiceUrl:   fmt.Sprintf("localhost:%d", server.Port()),
-		BatchLinger:  DefaultBatchLinger,
-		BatchMaxSize: 1,
-		BatchTimeout: DefaultBatchTimeout,
-	}
+	options := NewClientOptions(fmt.Sprintf("localhost:%d", server.Port())).
+		BatchMaxSize(1)
```
nahguam: e.g.

oxia/options.go
```diff
@@ -0,0 +1,48 @@
...
+		serviceUrl:   serviceUrl,
+		batchLinger:  defaultBatchLinger,
+		batchMaxSize: defaultBatchMaxSize,
+		batchTimeout: defaultBatchTimeout,
+	}
+}
+
+func (o ClientOptions) BatchLinger(batchLinger time.Duration) ClientOptions {
+	if batchLinger <= 0 {
+		panic("BatchLinger must be greater than zero")
```
nodece: Maybe we should avoid to use the `panic`.

nahguam: Why? It's not something that can be recovered from. In JVM terms it's the equivalent of a RuntimeException - a user coding / config error.

I could move the validation to a separate func and change the return of new client funcs to `(AsyncClient, error)`. Then we validate the options there?

something like:
```
func NewAsyncClient(options ClientOptions) (AsyncClient, error) {
    if err := options.validate(); err != nil {
        return nil, err
    }
    ...
}
```

nodece: The `panic` breaks the program running. Or you don't use the chain.

nodece: LGMT for your example.

merlimat: I think it's appropriate for invalid configurations to fail the process. 

I'd use `log.Fatal()` though to achieve the same result

oxia/options.go
```diff
@@ -0,0 +1,48 @@
...
+func (o ClientOptions) BatchMaxSize(batchMaxSize int) ClientOptions {
+	if batchMaxSize <= 0 {
+		panic("BatchMaxSize must be greater than zero")
+	}
+	o.batchMaxSize = batchMaxSize
+	return o
+}
+
+func (o ClientOptions) BatchTimeout(batchTimeout time.Duration) ClientOptions {
+	if batchTimeout <= 0 {
```
merlimat: 0 should be a valid value, meaning "no wait".

oxia/sync_client_impl.go
```diff
@@ -5,8 +5,7 @@ type syncClientImpl struct {
 }
 
 func NewSyncClient(options ClientOptions) SyncClient {
-	options.BatchMaxSize = 1
-	return newSyncClient(NewAsyncClient(options))
+	return newSyncClient(NewAsyncClient(options.BatchMaxSize(1)))
```
merlimat: The problem with sync-client is not the batching per se, but rather the extra wait time. 

We can have 2 classes of sync-client users: 
 1. Client that makes many sequential operations from the same go-routine. In this case, batching is useless and we want to send out each batch with 1 single request ASAP.
 2. Client that has many go-routines making concurrent calls. 
 
 In the 2nd scenario, there could be a lot of opportunities for batching.
 
 I'd say to have the batch-timeout set to 0 for sync-client. That would mean that a batch is triggered immediately. At the same time, when we trigger a batch, we should try to read all the requests that are available on the channel and pack them in the same batch. 
 
 This would adapt to support well both (1) and (2) scenarios for sync client.

oxia/options.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (o ClientOptions) BatchLinger(batchLinger time.Duration) ClientOptions {
+	if batchLinger <= 0 {
+		panic("BatchLinger must be greater than zero")
+	}
+	o.batchLinger = batchLinger
+	return o
+}
+
+func (o ClientOptions) BatchMaxSize(batchMaxSize int) ClientOptions {
```
merlimat: We should clarify that this is meaning "max requests per batch", rather than "max bytes size per batch"

oxia/options.go
nahguam: > The `panic` breaks the program running. Or you don't use the chain.

Yes, but it's unrecoverable so why force the `error` on the API?

The user can't reasonably do anything programmatically with the error. It indicates a misconfiguration which would mean restarting the program anyway.

nodece: > Yes, but it's unrecoverable so why force the `error` on the API?

It is better to return an `error` to the user, maybe this user doesn't care about this error.

Usually, this is reasonable, but the user must use `recover` to catch the `panic`, otherwise the program exits.

nahguam: I've been pondering this over the weekend. You're right. We can't anticipate what other things are going on in a user's application - they may need to gracefully close other resources. `error` is the right way to go.

oxia/options.go
nahguam: The timeout is for the receiving of the grpc response, I assume you mean for linger 👍

oxia/sync_client_impl.go
nahguam: Again, by `timeout`, I assume you mean `linger`.

I've now wired in so `BatchLinger=0` will send the batch immediately. However `MaxRequestsPerBatch=1` will also send immediately.

oxia/sync_client_impl.go
```diff
@@ -5,7 +5,8 @@ type syncClientImpl struct {
 }
 
 func NewSyncClient(options ClientOptions) SyncClient {
-	options.BatchMaxSize = 1
+	options.batchLinger = 0
+	options.maxRequestsPerBatch = 1
```
merlimat: What I was meaning is that we should keep `maxRequestsPerBatch` to default value of 100 or 1K, so that we can automatically cover the case where we have many go-routines using sync client.

PR #47, opened 2022-11-20
Repeatable read consistency by ReadBatch

andrasbeni:


server/kv/kv_pebble.go
```diff
@@ -231,6 +235,44 @@ func (p *PebbleSnapshotIterator) Value() ([]byte, error) {
 	return p.si.ValueAndErr()
 }
 
+/// Iterator wrapper methods for ReadBatch
+
+type PebbleReadBatch struct {
```
nahguam, line 240: Does this need to be exposed? It's being hidden behind an interface - `pebbleReadBatch`?

server/kv/db.go
```diff
@@ -100,22 +101,30 @@ func (d *db) addCommitIndex(commitIndex *proto.EntryId, batch WriteBatch) error
...
 	res := &proto.ReadResponse{}
 
-	//TODO Implement a ReadBatch to guarantee read consistency
-	kv := d.kv
+	rb := d.kv.NewReadBatch()
+	defer func() {
+		err := rb.Close()
+		if err != nil {
+			if e == nil {
+				r, e = nil, err
```
nahguam, line 112: Is this the right thing to do? This error is because the read batch couldn't close, but there may be a valid `ReadResponse`. Return the valid original response and the close error?

```suggestion
				e = err
```

What are the consequences of the failed close? Maybe it's a non-issue :)

server/kv/kv_pebble.go
andrasbeni: Probably not. I just blindly followed what's been done in this file (e.g. `PebbleIterator`)

server/kv/db.go
andrasbeni: The iterator holds some memory to be able to provide consistent reads. If it fails to close, it might cause a memory usage build-up. 
I have never seen go code that uses the other return values when an error is not nil, so I didn't even think of making it possible. I'd say if there's an issue with the server, let's just fail.

nahguam: Yep, let's just keep it simple

PR #54, opened 2022-11-21
Added quorum ack tracker

merlimat:
QuorumAckTracker has the logic for managing the commit and head indexes and blocking the go-routines waiting for these to be updated.

### DIFF
--- a/server/quorum_ack_tracker.go
+++ b/server/quorum_ack_tracker.go
@@ -65,7 +65,7 @@ type cursorAcker struct {
 func NewQuorumAckTracker(replicationFactor uint32, headIndex wal.EntryId) QuorumAckTracker {
 	q := &quorumAckTracker{
 		// Ack quorum is number of follower acks that are required to consider the entry fully committed
-		// We are suing RF/2 (and not RF/2 + 1) because the leader is already storing 1 copy locally
+		// We are using RF/2 (and not RF/2 + 1) because the leader is already storing 1 copy locally
 		requiredAcks:      replicationFactor / 2,
 		replicationFactor: replicationFactor,
 		headIndex:         headIndex,
--- a/server/util/bitset.go
+++ b/server/util/bitset.go
@@ -5,6 +5,8 @@ import (
 	"math/bits"
 )
 
+const MaxBitSetSize = 16
+
 // BitSet
 // Simplified and compact bitset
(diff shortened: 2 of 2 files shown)


server/quorum_ack_tracker.go
```diff
@@ -0,0 +1,169 @@
...
+
+type cursorAcker struct {
+	quorumTracker *quorumAckTracker
+	cursorIdx     int
+}
+
+func NewQuorumAckTracker(replicationFactor uint32, headIndex wal.EntryId) QuorumAckTracker {
+	q := &quorumAckTracker{
+		// Ack quorum is number of follower acks that are required to consider the entry fully committed
+		// We are suing RF/2 (and not RF/2 + 1) because the leader is already storing 1 copy locally
```
andrasbeni: nit: _using_

server/util/bitset.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"math/bits"
+)
+
+// BitSet
+// Simplified and compact bitset
+type BitSet struct {
+	bits uint16
```
andrasbeni, line 13: Can we please define a constant somewhere?
e.g. in this file `const MaxBitSetSize:= 16` and later in coordinator `const MaxReplicationFactor = util.MaxBitSetSize + 1`?

server/quorum_ack_tracker.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// WaitForHeadIndex
+	// Waits until the specified entry is written on the wal
+	WaitForHeadIndex(entryId wal.EntryId)
+
+	NewCursorAcker() (CursorAcker, error)
+}
+
+type quorumAckTracker struct {
+	sync.Mutex
```
andrasbeni, line 40: If we used RWMutex then `CommitIndex()` and `HeadIndex()` could make use of `RLock()`. Is this premature optimization?

merlimat: I'm trying to have the simplest solution for now. In practice, we might want to use atomic refs for these variables to avoid mutexes. That will also change when switching to use just the offset instead of `EntryId` for positioning in the log.

server/util/bitset.go
merlimat: 👍

PR #56, opened 2022-11-22
Follower cursor implementation

merlimat:
This is stacked on top of #54 

The `FollowerCursor` is created by the leader controller for each follower and it handles pushing the updates and receiving the acks back, passing them to the `AckQuorumTracker`.

### DIFF
--- a/server/wal/wal.go
+++ b/server/wal/wal.go
@@ -65,7 +65,7 @@ type WalReader interface {
 	ReadNext() (*proto.LogEntry, error)
 	// HasNext returns true if there is an entry to read.
 	// For a reverse WalReader this means the reader has not yet reached the beginning of the log.
-	// For a forward WalReader this means that we have not yet reached the offset that was the log end when the reader was created.
+	// For a forward WalReader this means that we have not yet reached the end of the wal
 	HasNext() bool
 }
 


server/wal/wal_tidwall.go
```diff
@@ -293,7 +293,8 @@ func (r *forwardReader) HasNext() bool {
 	if r.closed {
 		return false
 	}
-	return r.nextOffset <= r.maxIdInclusive.Offset && EntryIdFromProto(r.maxIdInclusive) != EntryId{}
+
+	return r.nextOffset <= r.wal.lastEntryId.Offset && EntryIdFromProto(r.maxIdInclusive) != EntryId{}
```
andrasbeni, line 297: Please update the documentation comment of `HasNext()` accordingly.

server/follower_cursor_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"testing"
+	"time"
+)
+
+func TestFollowerCursor(t *testing.T) {
+	var epoch uint64 = 1
+	var shard uint32 = 2
+
+	stream := newMockClientAddEntriesStream()
+	ackTracker := NewQuorumAckTracker(3, wal.EntryId{Epoch: 0, Offset: 0})
```
andrasbeni, line 16:
```suggestion
	ackTracker := NewQuorumAckTracker(3, wal.NonExistentEntryId)
```

andrasbeni: Maybe we should rename it to `wal.LogStartEntryId`?

server/wal/wal_tidwall.go
merlimat: The problem is that it's inconsistent with `ReadNext()`: 
 * ReadNext max offset is updated by reading the new value from the channel
 * HasNext is not updated

merlimat: I don't think we need to keep track of the readers and have this channel mechanism to notify them. Each reader can just read the variable of what is the latest entry committed in the wal.

server/follower_cursor_test.go
merlimat: I plan to use `int64` for offset anyway and use `-1` for all these scenarios

server/wal/wal_tidwall.go
merlimat: (updated the doc for now)

server/follower_cursor_test.go
andrasbeni: That makes sense.

server/wal/wal_tidwall.go
andrasbeni: Yes, I believe this series of pull requests will make a lot of what is there now redundant.

PR #57, opened 2022-11-22
Unify Wal testing

andrasbeni:


README.md
```diff
@@ -22,6 +22,10 @@ make test
 make docker
 ```
 
+### IDE settings
+Make sure to add `testing` tag to your compiler settings. This will add files with `//go:build testing` to the compile path.
+In GoLand you can do that by adding `testing` in `Preferences`/`Go`/`Build Tags & Vendoring`/`Custom tags` *and* adding `--tags testing` to `Run`/`Edit Configurations...`/`Edit configuration templates`/`Go Test`/`Go tool arguments`.
```
merlimat: Instead of requiring these changes here, I'd prefer to just put the memory wal as part of the production code, just not used. There should be no harm in that.

andrasbeni: Done

server/wal/wal_memory.go
```diff
@@ -202,8 +203,24 @@ func (w *inMemoryWal) Append(entry *proto.LogEntry) error {
...
+	if len(w.log) != 0 {
+		lastEntryId = w.log[len(w.log)-1].EntryId
+	}
+
+	lastEpoch := lastEntryId.Epoch
+	lastOffset := lastEntryId.Offset
+	nextEpoch := entryId.Epoch
+	nextOffset := entryId.Offset
+	if (lastOffset == 0 && lastEpoch == 0 && !(nextEpoch == 1 && nextOffset == 0)) ||
+		(lastEpoch > 0 && ((nextOffset != lastOffset+1) || (nextEpoch < lastEpoch))) {
```
nahguam, line 217: that's a mouthful!

server/wal/wal_test.go
```diff
@@ -10,9 +10,59 @@ import (
...
+}
+
+func (_ *inMemoryWalFactoryFactory) Persistent() bool {
+	return false
+}
+
+func (_ *tidwallWalFactoryFactory) NewWalFactory(t *testing.T) WalFactory {
 	dir := t.TempDir()
 	f := NewWalFactory(&WalFactoryOptions{dir})
+	return f
```
nahguam, line 37:
```suggestion
  return NewWalFactory(&WalFactoryOptions{dir})
```

server/wal/wal_memory.go
andrasbeni: Matteo has a change planned where he moves from unsigned int to signed for epoch and offset. This expression will be much simpler with the ability to of using offset=-1 instead of reserving (epoch=0&offset=0) for "No entry".

PR #58, opened 2022-11-22
Refactor leader changes

merlimat:
This is stacked on top of #56 

Refactored controller for handling "leader" logic.

Added mocked RPC client for testing - More failure test scenario to add later.


server/follower_controller.go
```diff
@@ -31,7 +31,7 @@ type FollowerController interface {
 	//
 	// When a node is fenced it cannot:
 	// - accept any writes from a client.
-	// - accept add entry requests from a leader.
+	// - accept add entry addEntryRequests from a leader.
```
andrasbeni:
```suggestion
	// - accept addEntryRequests from a leader.
```

proto/coordination.proto
```diff
@@ -8,8 +8,6 @@ option go_package = "github.com/streamnative/oxia/proto";
 service OxiaControl {
   rpc Fence(FenceRequest) returns (FenceResponse);
   rpc BecomeLeader(BecomeLeaderRequest) returns (BecomeLeaderResponse) {}
-  rpc AddFollower(AddFollowerRequest) returns (CoordinationEmpty) {}
```
andrasbeni, line 11: Can you please explain the motivation behind the removal of this method? If I understand correctly, AddFollower is used by the controller to add followers to the ensemble* that did not respond to the Fence request until the quorum has. Will the controller wait for all followers to respond before sending out the BecomeLeader request? If so, will it not impact availability negatively? 


*I use ensemble as the group of servers that participate in replicating a given shard (leader + followers). Do we have a name for this?

proto/coordination.proto
```diff
@@ -31,7 +24,6 @@ message EntryId {
 message LogEntry {
   EntryId entry_id = 1;
   bytes value = 2;
-  uint64 timestamp = 3;
 }
 
 message CoordinationEmpty {}
```
andrasbeni: If we decide to get rid of AddFollower, we don't need CoordinationEmpty either.

server/leader_controller.go
```diff
@@ -0,0 +1,333 @@
...
+// with its head index.
+//
+// When a node is fenced it cannot:
+//   - accept any writes from a client.
+//   - accept add entry addEntryRequests from a leader.
+//   - send any entries to followers if it was a leader.
+//
+// Any existing follow cursors are destroyed as is any state
+// regarding reconfigurations.
+func (lc *leaderController) Fence(req *proto.FenceRequest) (*proto.FenceResponse, error) {
```
andrasbeni, line 113: This is never called because you change to follower controller whenever a Fence request hits the internal rpc server.

proto/coordination.proto
merlimat: `AddFollower` is only used in the context of node swap. In regular leader change, the coordinator passes the complete list of followers in the `BecomeLeader` request

server/leader_controller.go
merlimat: That's a good point. 
I believe it should be safe to switch to Follower after a Fence, provided that the Fence will only come at the leader election time, so we shouldn't be getting this and flip-flop from Leader -> Follower -> Leader again.

Another important part that is missing in handling the fencing is that we must persist the fenced status, otherwise we might crash and come back accepting entries from a previous epoch.

proto/coordination.proto
andrasbeni: That's not how I read the specification (https://github.com/streamnative/oxia/blob/main/tlaplus/OxiaReplication.tla#L539-L546, https://github.com/streamnative/oxia/blob/main/tlaplus/OxiaReplication.tla#L698-L709). But if that's the case, let's remove AddFollower from the spec also.

server/leader_controller.go
andrasbeni: >I believe it should be safe to switch to Follower after a Fence, provided that the Fence will only come at the leader election time, so we shouldn't be getting this and flip-flop from Leader -> Follower -> Leader again.

What happens when we change one of the followers of the shard? I believe that needs fencing (because there's no other way to remove a probably dead follower) yet the leader stays the same. 

>Another important part that is missing in handling the fencing is that we must persist the fenced status, otherwise we might crash and come back accepting entries from a previous epoch.

I don't think that's a problem. Let's say epoch n has ended, and we got a fencing request with epoch n+1 and crashed before responding to the request. We might or might not have truncated the log, it's not important. Until the coordinator realizes the issue, we can accept entries from the old leader, because when the new leader learns about this follower, it will send a truncate request with epoch n+1 that removes all non-committed entries from epoch n. Sure, it's a waste of network resources for a few hundred milliseconds, but I would not worry about it.

merlimat: @andrasbeni I've changed to fence either the follower or leader, whatever is currently loaded. This won't change any of the intended logic.

merlimat: >I don't think that's a problem. Let's say epoch n has ended, and we got a fencing request with epoch n+1 and crashed before responding to the request. We might or might not have truncated the log, it's not important. Until the coordinator realizes the issue, we can accept entries from the old leader, because when the new leader learns about this follower, it will send a truncate request with epoch n+1 that removes all non-committed entries from epoch n. Sure, it's a waste of network resources for a few hundred milliseconds, but I would not worry about it.

 The problem is that there could be an old-leader at epoch n-1. By receiving the "fence" the follower (along with the rest of the majority) is moving into epoch n. 
 
If the follower crashes and comes back in epoch n-1, it will still be able to accept writes for that epoch, even after a truncate.

The fence status must be persistent to avoid all these cases. That's also the case in the spec where the log and local state of each node is not reset after a crash.

andrasbeni: I think I imagined the system even less fault tolerant than you do. In my mind, if the node goes down, it will respond with an error to the subsequent request that contradicts what it knows about the world, and eventually, the controller starts a new epoch. E.g., if it goes down after a Fence request but comes back by the Truncate request, it responds with an error. 
But I can live with the status persisted in the key-value store like commitIndex is, if it helps recovery.

PR #60, opened 2022-11-23
[Feature][CLI] Sketch out commands for client

teabot:
# Motivation
To be able to easily CRUD oxia records from the command line, and exercise oxia as a programmatic client might.

# Changes
* Sketched out the different commands and arguments — no linkage to actual client yet


cli/client/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+			fmt.Printf("Delete %v [expectedVersion=%v]\n", args[0], args[1])
+		} else {
+			fmt.Printf("Delete %v\n", args[0])
+		}
+	},
+}
+var deleteRange = &cobra.Command{
+	Use:   "delete-range [key-start] [key-end]",
+	Short: "Deletes all records within key range",
+	Long:  `Deletes all records with keys within the range [key-start:key-end). Assumes a UTF-8 byte sort order on the keys`,
```
andrasbeni: I believe we don't actually use lexicographical ordering. See `CompareWithSlashes` in kv_pebble.go.

cli/client/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+	RootCmd.AddCommand(get)
+	RootCmd.AddCommand(getRange)
+	RootCmd.AddCommand(shardAssignments)
+	RootCmd.AddCommand(batchPut)
+	RootCmd.AddCommand(batchDelete)
+	RootCmd.AddCommand(batchGet)
+}
+
+var put = &cobra.Command{
+	Use:   "put [key] [value] [expected-version]",
```
andrasbeni: I'd prefer expected-version as a flag like `--expected-version 7` or `-v 7`. I could even imagine key and value being a flag, but wouldn't say I prefer that. 

Same for all commands below.

cli/client/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+
+	RootCmd = &cobra.Command{
+		Use:   "client",
+		Short: "Mutations of records",
+		Long:  `Operations to get, create, delete, and modify key-value records in an oxia cluster`,
+	}
+)
+
+func init() {
+	RootCmd.Flags().StringVarP(&serviceAddr, "service-address", "a", "localhost:9190", "Service address")
```
nahguam: > `serviceAddr`

👍  We should also change `oxia.ClientOptions.serviceUrl` to `serviceAddr`

cli/client/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+	RootCmd.Flags().Int64Var(&batchLinger, "batch-linger", int64(oxia.DefaultBatchLinger), "Batch linger in milliseconds")
+	RootCmd.Flags().IntVar(&maxRequestsPerBatch, "max-requests-per-batch", oxia.DefaultMaxRequestsPerBatch, "Maximum requests per batch")
+	RootCmd.Flags().Int64Var(&batchRequestTimeout, "batch-request-timeout", int64(oxia.DefaultBatchRequestTimeout), "Batch timeout in seconds")
+
+	RootCmd.AddCommand(put)
+	RootCmd.AddCommand(deleteCmd)
```
nahguam: > `deleteCmd`

Is this because `delete` would collide with the builtin function?

cli/client/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+		Short: "Mutations of records",
+		Long:  `Operations to get, create, delete, and modify key-value records in an oxia cluster`,
+	}
+)
+
+func init() {
+	RootCmd.Flags().StringVarP(&serviceAddr, "service-address", "a", "localhost:9190", "Service address")
+	RootCmd.Flags().Int64Var(&batchLinger, "batch-linger", int64(oxia.DefaultBatchLinger), "Batch linger in milliseconds")
+	RootCmd.Flags().IntVar(&maxRequestsPerBatch, "max-requests-per-batch", oxia.DefaultMaxRequestsPerBatch, "Maximum requests per batch")
+	RootCmd.Flags().Int64Var(&batchRequestTimeout, "batch-request-timeout", int64(oxia.DefaultBatchRequestTimeout), "Batch timeout in seconds")
```
nahguam: `int64(DefaultBatchRequestTimeout)` will print it in nanos. To match the description it'd be `int64(DefaultBatchRequestTimeout.Seconds())`

cli/client/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+
+var (
+	serviceAddr         string
+	batchLinger         int64
+	maxRequestsPerBatch int
+	batchRequestTimeout int64
+
+	RootCmd = &cobra.Command{
+		Use:   "client",
+		Short: "Mutations of records",
```
nahguam: not just mutations

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM golang:1.19-alpine as build
 RUN apk add --no-cache protobuf make git
 
 RUN go install github.com/golang/protobuf/protoc-gen-go@latest
-RUN go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+RUN go install google.golang.org/grpc/cli/protoc-gen-go-grpc@latest
```
merlimat: I think we shouldn't change these urls or we wan't be able to fetch the tools :)

Makefile
```diff
@@ -1,7 +1,7 @@
 
 .PHONY: build
 build: proto
-	go build -v -o bin/oxia ./cmd
+	go build -v -o bin/oxia ./cli
```
merlimat: As shown by all these other tools `cmd` is kind of the convention in Go for the package where to put CLI tools.

cli/client/cmd.go
merlimat: 👍 

Also if the value contains spaces, new-lines or binary stuff, we should give the option to either: 
 * Read from a file
 * Read from stdin

cli/client/cmd.go
```diff
@@ -0,0 +1,139 @@
...
+	Use:   "get-range [key-start] [key-end]",
+	Short: "Gets all records within key range",
+	Long:  `Gets all records with keys within the range [key-start:key-end). Assumes a UTF-8 byte sort order on the keys`,
+	Args:  cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Printf("Get range %v -> %v\n", args[0], args[1])
+	},
+}
+var batchPut = &cobra.Command{
+	Use:   "batch-put [csv-file]",
```
merlimat: We don't really have the concept of "batch" in the API. It's really an implementation detail of how things gets packed up for efficiency.

We could offer to insert "many" keys, though let's not call it a batch :D

cli/main.go
```diff
@@ -20,14 +17,9 @@ var (
...
 
 func init() {
-	//rootCmd.PersistentFlags().StringVarP(&userLicense, "license", "l", "", "name of license for the project")
 	rootCmd.PersistentFlags().BoolVarP(&common.LogDebug, "log-debug", "d", false, "Enable debug logs")
 	rootCmd.PersistentFlags().BoolVarP(&common.LogJson, "log-json", "j", false, "Print logs in JSON format")
-
-	rootCmd.AddCommand(server.Cmd)
-	rootCmd.AddCommand(operator.Cmd)
-	rootCmd.AddCommand(client.Cmd)
-	rootCmd.AddCommand(standalone.Cmd)
```
merlimat: The root command is already used to start the server/operator/standalone components, we shouldn't remove them

cli/client/cmd.go
nahguam: My fault, I said it was this as that was what we discussed in a sync meeting. We also said the comparer would be pluggable (I created a [ticket](https://www.notion.so/streamnativeio/a8a710bff405448b911b45c2349ebdf3?v=1c376b03b96d4293ab5f84dcd735973f&p=1ccdb7c8e6fc429f9ac4b635a80c8ea8&pm=s))

nahguam: We should [document](https://www.notion.so/streamnativeio/a8a710bff405448b911b45c2349ebdf3?v=1c376b03b96d4293ab5f84dcd735973f&p=18b68626bac145be8d1d1264467212f5&pm=s) the `CompareWithSlashes` comparer

cli/client/cmd.go
nahguam: https://github.com/streamnative/oxia/pull/64

cli/client/cmd.go
nahguam: That's true, it's overloading the word. `bulk`?

cli/client/cmd.go
teabot: ok, good catch. I'll update the usage message.

cli/client/cmd.go
teabot: Ok, I struggled with this decision myself. I'd say that for mandatory input we should go with an arg, but for selectable input we should used flags. So in this example, I think key should be an arg, but expectedVersion and value should be flags. This pattern should then be applied throughout.

cli/client/cmd.go
teabot: Approved

cli/client/cmd.go
teabot: correct

cli/client/cmd.go
teabot: Ah yes, I see that there is a problem. So `batchRequestTimeout` is a `time.Duration` (`i64` with `ns` resolution), but the input will be a numeric string with `s` resolution. I need to have a look where I can elegantly perform the conversion with cobra.

cli/client/cmd.go
teabot: What would be a good short form that isn't the long form?

Dockerfile
teabot: Hah, good catch. Thank you.

Makefile
teabot: Noted

cli/client/cmd.go
teabot: Perhaps let's revisit the flags on `put`, `get`, `delete` so that can on operate both on single keys and sets of keys without the need for additional commands,

cli/main.go
teabot: Ok, I can reinstate them. I was under the impression that these weren't the active implementations.

merlimat: * `server` starts fine, though not fully functioning at this point since it needs input from the coordinator.
 * `operator` is just a test thing now, though we should leave it and later add the real stuff
 *  `client` was meant to be this CLI client :) 
 *  `standalone` is already functional 😎! You can use it to test the CLI

Dockerfile
teabot: Done

Makefile
teabot: Done

cli/client/cmd.go
teabot: > Perhaps let's revisit the flags on `put`, `get`, `delete` so that can on operate both on single keys and sets of keys without the need for additional commands,

This is what I did in the end - I think it's much simpler.

cli/client/cmd.go
teabot: I've excluded any mention of sort order currently.

cli/main.go
teabot: Reinstated

cli/client/cmd.go
teabot: Done

PR #62, opened 2022-11-23
Assignment dispatcher

andrasbeni:


proto/coordination.proto
```diff
@@ -6,6 +6,8 @@ option go_package = "github.com/streamnative/oxia/proto";
 
 // controller -> node
 service OxiaControl {
+  rpc ShardAssignment(stream CoordinationShardAssignmentRequest) returns (stream CoordinationShardAssignmentResponse) {}
```
merlimat: There should be no need to stream back a set of responses. 

This is essentially a notification channel and the coordinator doesn't need to get any confirmation or any other info back.

proto/coordination.proto
```diff
@@ -94,3 +96,49 @@ message AddEntryResponse {
   EntryId entry_id = 2;
   bool invalid_epoch = 3;
 }
+
+// Shard assignment
+
+message CoordinationShardAssignmentRequest {
```
merlimat: I would rather reuse the existing public API definitions because otherwise, we're just duplicating the same data structures and we need to translate between them which is tedious.

If in the future we get into the need to add more information that it's only server specific, we can always add a new endpoint for that, though that's not something we should be worrying about right now.

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+func (s *shardAssignmentDispatcher) Initialized() bool {
+	s.Lock()
+	defer s.Unlock()
+	return s.initialized
+}
+
+func (s *shardAssignmentDispatcher) ShardAssignment(srv proto.OxiaControl_ShardAssignmentServer) error {
+	for {
+		request, err := srv.Recv()
+		if err == nil {
```
merlimat: request can also be `nil` when the other side is closing the stream

server/internal_rpc_server.go
```diff
@@ -13,21 +13,27 @@ import (
 )
 
 const (
-	metadataShardId = "shard-id"
+	metadataShardId = "shardAssignment-id"
```
merlimat: Why do we need to change all references of `shard` and `shard-id` into `shardAssignement` ?

proto/coordination.proto
andrasbeni: Would you like it to be moved to a common.proto that is imported by both or should coordination.proto import client.proto?

proto/coordination.proto
andrasbeni: Removed response stream.

proto/coordination.proto
andrasbeni: Imported from client.proto. We can move it to common later if we want to.

server/assignment_dispatcher.go
andrasbeni: Thanks, I wasn't aware of it

server/internal_rpc_server.go
andrasbeni: Thanks for pointing this out. I don't really know how I did this. But not intentionally.

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+	assignments    map[uint32]shardAssignment
+	clients        []Client
+	stopRecv       context.CancelFunc
+}
+
+func (s *shardAssignmentDispatcher) AddClient(clientStream Client) error {
+	s.Lock()
+	defer s.Unlock()
+	if !s.initialized {
+		return errors.New("oxia: server not initialized yet")
```
merlimat: Let's use constant for these errors, so that it becomes easier to test

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+
+func removeAt[T any](slice []T, idx int) []T {
+	if idx == len(slice)-1 {
+		return slice[:idx]
+	} else {
+		return append(slice[:idx], slice[idx+1:]...)
+	}
+}
+
+func (s *shardAssignmentDispatcher) convertAssignments() *proto.ShardAssignmentsResponse {
```
merlimat, line 135: Why do we need this conversion? Couldn't we just result `s.assignments` as a `*proto.ShardAssignmentsResponse`?

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return result
+}
+
+func NewShardAssignmentDispatcher() ShardAssignmentsDispatcher {
+	return &shardAssignmentDispatcher{
+		initialized:    false,
+		closed:         false,
+		shardKeyRouter: proto.ShardKeyRouter_UNKNOWN,
+		assignments:    make(map[uint32]shardAssignment, 1000),
```
merlimat: Why convert into a map?

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+
+	s.initialized = true
+	return nil
+}
+
+func removeAt[T any](slice []T, idx int) []T {
+	if idx == len(slice)-1 {
+		return slice[:idx]
+	} else {
+		return append(slice[:idx], slice[idx+1:]...)
```
merlimat: Maybe a map would be more appropriate to remove clients from arbitrary positions. We could assign a sequential int64 to each client when they're are added.

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+func NewStandaloneShardAssignmentDispatcher(address string, numShards uint32) ShardAssignmentsDispatcher {
+	assignmentDispatcher := &shardAssignmentDispatcher{
+		initialized:    false,
+		closed:         false,
+		shardKeyRouter: proto.ShardKeyRouter_UNKNOWN,
+		assignments:    make(map[uint32]shardAssignment, 1000),
+		clients:        make([]Client, 0, 1000),
```
merlimat: I'd start with default array size

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,210 @@
...
+		current.lowerBoundInclusive = assignment.GetInt32HashRange().MinHashInclusive
+		current.upperBoundExclusive = assignment.GetInt32HashRange().MaxHashExclusive
+		s.assignments[shard] = current
+	}
+
+	for idx, client := range s.clients {
+		err := client.Send(request)
+		if err != nil {
+			s.clients = removeAt(s.clients, idx)
+			return err
```
merlimat: If we fail to update 1 client, we shouldn't be failing here for all the other clients. 

We should just close the stream and return error to that particular client

andrasbeni: How can we close the stream?

server/assignment_dispatcher.go
andrasbeni: Using a map with default size

server/assignment_dispatcher.go
andrasbeni: I want to allow non-continuous shard IDs that will be useful when shard splitting is implemented.

proto/coordination.proto
```diff
@@ -4,8 +4,12 @@ package coordination;
 
 option go_package = "github.com/streamnative/oxia/proto";
 
+import "proto/client.proto";
+
 // controller -> node
 service OxiaControl {
+  rpc ShardAssignment(stream io.streamnative.oxia.proto.ShardAssignmentsResponse) returns (CoordinationEmpty) {}
```
nahguam: I thought we had settled on the convention of:

```
rpc Thing(ThingRequest) returns (ThingResponse)
```

i.e. each rpc has a dedicated request/response type so we can gracefully evolve the api. Reusing the `*Response` message from the client as a request and `CoordinationEmpty` doesn't conform :)

Also, in `ShardAssignmentsResponse` it has `ShardKeyRouter` which does not need to be sent on every stream send. It only needs to be sent once. I don't think we have any intention of supporting anything other than key hashing (xxhash3) so should we remove this? (slightly off topic, I know)

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,193 @@
...
+	defer s.Unlock()
+	if !s.initialized {
+		return ErrorNotInitialized
+	}
+	assignments := s.convertAssignments()
+	err := clientStream.Send(assignments)
+	if err != nil {
+		return err
+	}
+	s.clients[s.nextClientId] = clientStream
```
nahguam, line 52: Does `nextClientId` get incremented?

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+type shardAssignmentDispatcher struct {
+	sync.Mutex
+	initialized    bool
+	closed         bool
+	shardKeyRouter proto.ShardKeyRouter
+	assignments    map[uint32]*proto.ShardAssignment
+	clients        map[uint32]Client
+	nextClientId   uint32
```
nahguam: Does it need to be `uint32` or could it just be `int` (less cognitive load on the reader)

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+	s.clients[s.nextClientId] = clientStream
+	return nil
+}
+
+func (s *shardAssignmentDispatcher) Close() error {
+	s.Lock()
+	defer s.Unlock()
+	s.closed = true
```
nahguam: Is this needed? It doesn't seem to be referenced

server/assignment_dispatcher.go
```diff
@@ -0,0 +1,193 @@
...
+
+type Client interface {
+	Send(*proto.ShardAssignmentsResponse) error
+}
+
+type ShardAssignmentsDispatcher interface {
+	io.Closer
+	Initialized() bool
+	ShardAssignment(proto.OxiaControl_ShardAssignmentServer) error
+	AddClient(Client) error
```
nahguam, line 21: Do we need to remove at some point, when the client disconnects and/or the session expires?

proto/coordination.proto
andrasbeni: * I was going to define it differently, but @merlimat suggested to reuse what we have in this comment: https://github.com/streamnative/oxia/pull/62#discussion_r1030941319
* I don't mind removing the shard key router, but let's do it in a separate PR.

server/assignment_dispatcher.go
andrasbeni: We don't need to do that from the outside. In `updateShardAssignment` we remove clients that produce errors.

server/assignment_dispatcher.go
andrasbeni: The unspecificity of `int` puts more cognitive load on my odd brain than the well-specified `uint32`. But I'm not going to be the only one supporting this, so I changed it.

server/assignment_dispatcher.go
andrasbeni: Should have got incremented, now it is fixed

server/assignment_dispatcher.go
andrasbeni: You're right, it's not used.

server/assignment_dispatcher.go
merlimat: I think server can just pass through the same `ShardAssignmentsResponse` object, without having to interpret it or modify it. 

eg: 

```diff
diff --git server/assignment_dispatcher.go server/assignment_dispatcher.go
index 255619f..429306c 100644
--- server/assignment_dispatcher.go
+++ server/assignment_dispatcher.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc/peer"
 	"io"
 	"math"
 	"oxia/proto"
@@ -12,20 +13,22 @@ import (
 
 type Client interface {
 	Send(*proto.ShardAssignmentsResponse) error
+
+	Context() context.Context
 }
 
 type ShardAssignmentsDispatcher interface {
 	io.Closer
 	Initialized() bool
-	ShardAssignment(proto.OxiaControl_ShardAssignmentServer) error
-	AddClient(Client) error
+	ShardAssignment(stream proto.OxiaControl_ShardAssignmentServer) error
+	AddClient(client Client) error
 }
 
 type shardAssignmentDispatcher struct {
 	sync.Mutex
 	initialized    bool
 	shardKeyRouter proto.ShardKeyRouter
-	assignments    map[uint32]*proto.ShardAssignment
+	assignments    *proto.ShardAssignmentsResponse
 	clients        map[int]Client
 	nextClientId   int
 	stopRecv       context.CancelFunc
@@ -33,9 +36,6 @@ type shardAssignmentDispatcher struct {
 
 var (
 	ErrorNotInitialized = errors.New("oxia: server not initialized yet")
-	ErrorUnknownRouter  = errors.New("oxia: unknown shard key router")
-	ErrorChangedRouter  = errors.New("oxia: changing shard key router is not supported")
-	ErrorShardSplitting = errors.New("oxia: shard splitting is not yet supported")
 )
 
 func (s *shardAssignmentDispatcher) AddClient(clientStream Client) error {
@@ -44,8 +44,7 @@ func (s *shardAssignmentDispatcher) AddClient(clientStream Client) error {
 	if !s.initialized {
 		return ErrorNotInitialized
 	}
-	assignments := s.convertAssignments()
-	err := clientStream.Send(assignments)
+	err := clientStream.Send(s.assignments)
 	if err != nil {
 		return err
 	}
@@ -97,33 +96,19 @@ func (s *shardAssignmentDispatcher) ShardAssignment(srv proto.OxiaControl_ShardA
 	}
 }
 
-func (s *shardAssignmentDispatcher) updateShardAssignment(request *proto.ShardAssignmentsResponse) error {
+func (s *shardAssignmentDispatcher) updateShardAssignment(assignments *proto.ShardAssignmentsResponse) error {
 	s.Lock()
 	defer s.Unlock()
-	if request.ShardKeyRouter == proto.ShardKeyRouter_UNKNOWN {
-		return ErrorUnknownRouter
-	}
-
-	if !s.initialized {
-		s.shardKeyRouter = request.ShardKeyRouter
 
-	}
-	if s.shardKeyRouter != request.ShardKeyRouter {
-		return ErrorChangedRouter
-	}
-	for _, assignment := range request.Assignments {
-		shard := assignment.ShardId
-		_, found := s.assignments[shard]
-		if s.initialized && !found {
-			return ErrorShardSplitting
-		}
-		s.assignments[shard] = assignment
-	}
+	s.assignments = assignments
 
 	for id, client := range s.clients {
-		err := client.Send(request)
+		err := client.Send(assignments)
 		if err != nil {
-			log.Err(err).Msg("Error sending shard assignment update to client")
+			peer, _ := peer.FromContext(client.Context())
+			log.Warn().Err(err).
+				Interface("client", peer.Addr.String()).
+				Msg("Failed to send shard assignment update to client")
 			delete(s.clients, id)
 		}
 	}
@@ -132,33 +117,19 @@ func (s *shardAssignmentDispatcher) updateShardAssignment(request *proto.ShardAs
 	return nil
 }
 
-func (s *shardAssignmentDispatcher) convertAssignments() *proto.ShardAssignmentsResponse {
-	assignments := make([]*proto.ShardAssignment, 0, len(s.assignments))
-	for _, assignment := range s.assignments {
-		assignments = append(assignments, assignment)
-	}
-	result := &proto.ShardAssignmentsResponse{
-		Assignments:    assignments,
-		ShardKeyRouter: proto.ShardKeyRouter(s.shardKeyRouter.Number()),
-	}
-	return result
-}
-
 func NewShardAssignmentDispatcher() ShardAssignmentsDispatcher {
 	return &shardAssignmentDispatcher{
-		initialized:    false,
-		shardKeyRouter: proto.ShardKeyRouter_UNKNOWN,
-		assignments:    make(map[uint32]*proto.ShardAssignment),
-		clients:        make(map[int]Client),
+		initialized: false,
+		assignments: nil,
+		clients:     make(map[int]Client),
 	}
 }
 
 func NewStandaloneShardAssignmentDispatcher(address string, numShards uint32) ShardAssignmentsDispatcher {
 	assignmentDispatcher := &shardAssignmentDispatcher{
-		initialized:    false,
-		shardKeyRouter: proto.ShardKeyRouter_UNKNOWN,
-		assignments:    make(map[uint32]*proto.ShardAssignment),
-		clients:        make(map[int]Client),
+		initialized: false,
+		assignments: nil,
+		clients:     make(map[int]Client),
 	}
 	res := &proto.ShardAssignmentsResponse{
 		ShardKeyRouter: proto.ShardKeyRouter_XXHASH3,

```

merlimat: I think the part that was not clear is that each "assignement" brings **all** the shards information. 

That makes it easier for client implementation that they don't need to do a lot of complex logic.

andrasbeni: Oh, I did not think of that. Thanks

PR #67, opened 2022-11-28
Use buffered channel to improve batcher

### LINKED ISSUE #43: Support back pressure in a performant way in `batch.Batcher`
The main event loop in `batch.Batcher.run()` completes (sends gRPC requests) synchronously. This means it blocks the event loop, which in turn blocks the input channel `callC` which is unbuffered. This results in the main `oxia.Client` methods blocking.

We could buffer the `callC` channel to allow some records to continue to be inserted (up to a limit) while the current batch is being completed. And/or we could complete the batch in a goroutine with some kind of barrier to prevent ending up with no back pressure at all.

nodece:
Fix #43 

Use buffered channel to improve batcher, the default buffered size equals the CPU number. 

### Changes

- Add buffered size flag to the client cli
- Add buffered size to the client option
- Move the `BatcherFactory` to `batcher_factory.go` from `batcher.go`
- Make the `callC` as buffered channel

### DIFF
--- a/server/wal/wal.go
+++ b/server/wal/wal.go
@@ -18,7 +18,7 @@ type WalFactoryOptions struct {
 }
 
 var DefaultWalFactoryOptions = &WalFactoryOptions{
-	LogDir: "wal",
+	LogDir: "data/wal",
 }
 
 type WalFactory interface {
--- a/server/wal/wal_tidwall.go
+++ b/server/wal/wal_tidwall.go
@@ -5,8 +5,8 @@ import (
 	"github.com/pkg/errors"
 	tidwall "github.com/tidwall/wal"
 	pb "google.golang.org/protobuf/proto"
-	"os"
 	"oxia/proto"
+	"path/filepath"
 	"sync"
 )
 
@@ -38,7 +38,8 @@ type tidwallWal struct {
 
 func newTidwallWal(shard uint32, dir string) (Wal, error) {
 	opts := tidwall.DefaultOptions
-	log, err := tidwall.Open(fmt.Sprintf("%s%c%06d", dir, os.PathSeparator, shard), opts)
+	walPath := filepath.Join(dir, fmt.Sprint("shard-", shard))
+	log, err := tidwall.Open(walPath, opts)
 	if err != nil {
 		return nil, err
 	}


oxia/internal/batch/batcher.go
```diff
@@ -93,9 +55,11 @@ func (b *batcherImpl) run() {
 				batch = nil
 			}
 		case <-timeout:
-			timer.Stop()
-			batch.Complete()
-			batch = nil
+			if batch != nil {
+				timer.Stop()
+				batch.Complete()
```
nahguam, line 60: Now there are more threads on this event loop, there's a race here (and in close) with the main call. 1 thread might be completing (or failing) a batch while another is still trying to add more calls. We'd probably need to add some locking here.

nodece: > 1 thread might be completing (or failing) a batch while another is still trying to add more calls.

Do you mean we must wait for a batch to finish, then handle the next batch?

Right now there are multiple goroutines to read the `callC`, each batch has its goroutine, and other goroutine doesn't handle the `batch` of dequeue.

nahguam: Actually, I see now it'll be processing multiple batches concurrently. I'm not sure that's the right approach. I think the right approach would be to handle the batch backpressure on the existing single goroutine properly. i.e. allow the next batch to start being built while the current one is being completed. But ensuring we don't build a queue of waiting batches.

nodece: Sounds good. 
> But ensuring we don't build a queue of waiting batches.

I think we can use the buffered channel, we can receive multiple batch requests, and use one goroutine to handle the batch requests, but there have a lot of requests needed to handle.

nahguam: Yes, and the full buffer blocking the client should be enough to control the back pressure.

On a related note, during a discussion, @merlimat suggested that with a buffered channel and linger=0, it could drain the channel before completing the batch.

nodece: So do you want to remove the `linger`? That looks good to me.

nahguam: No, leave linger in. I'm saying if the channel is now buffered, when the batcher creates a new batch and `linger=0` there may be more than 1 message waiting. The current logic will just take 1 message and send immediately send a batch. Instead it should drain all the buffered messages and send the batch.

oxia/internal/batch/batcher_factory.go
```diff
@@ -0,0 +1,44 @@
...
+	"oxia/oxia/internal"
+	"oxia/oxia/internal/metrics"
+	"time"
+)
+
+type BatcherFactory struct {
+	Executor            internal.Executor
+	Linger              time.Duration
+	MaxRequestsPerBatch int
+	MaxJobs             int
```
nahguam: This is now the channel buffer size, so perhaps a better name would be `BatcherBufferSize`?

oxia/internal/batch/batcher.go
```diff
@@ -93,9 +55,11 @@ func (b *batcherImpl) run() {
 				batch = nil
 			}
 		case <-timeout:
-			timer.Stop()
-			batch.Complete()
-			batch = nil
+			if batch != nil {
```
nahguam, line 58: Shouldn't be necessary as the timer will only fire if we've received at least one call to start a new batch but doesn't do any harm being here.

oxia/internal/batch/batcher_factory.go
nodece: LGTM

oxia/internal/batch/batcher.go
nodece: No, sometimes there is nil.

PR #72, opened 2022-11-28
Introduce Container to manage grpcServer & listener lifecycle

nahguam:


server/container/container.go
```diff
@@ -0,0 +1,58 @@
...
+	c.log = log.With().
+		Str("container", name).
+		Str("bindAddress", listener.Addr().String()).
+		Logger()
+
+	go func() {
+		if err := c.server.Serve(listener); err != nil {
+			c.log.Fatal().Err(err).Msg("Failed to start serving")
+		}
+		fmt.Println("XXX")
```
andrasbeni:
```suggestion
```

server/container/container.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func Start(name string, port int, registerFunc func(grpc.ServiceRegistrar)) (*Container, error) {
+	c := &Container{
+		server: grpc.NewServer(),
+	}
+
+	registerFunc(c.server)
+
+	listener, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
```
andrasbeni, line 26: As I can see you are creating a new listener every time you call Start. But you call Start from both internal and public rpc server code, so there will be a conflict.

nahguam: Public and internal are configured to run on different ports, so there are 2 listeners.

nahguam: Would it be clearer if the function was `StartNew`?

andrasbeni: It was clear enough. What wasn't clear is that we use different ports.

PR #75, opened 2022-11-29
Standalone implementation using leader controller

merlimat:
Oxia standalone implementation should use the leader controller so that it's very similar to the real server. 

All the writes are written to the WAL and replication factor is set to 1

### DIFF
--- a/standalone/standalone_rpc_server.go
+++ b/standalone/standalone_rpc_server.go
@@ -42,20 +42,6 @@ func NewStandaloneRpcServer(port int, advertisedPublicAddress string, numShards 
 		log: log.With().
 			Str("component", "standalone-rpc-server").
 			Logger(),
-		//=======
-		//	identityAddr         string
-		//	numShards            uint32
-		//	dbs                  map[uint32]kv.DB
-		//	Container            *container.Container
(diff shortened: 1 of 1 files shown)


standalone/standalone_rpc_server.go
```diff
@@ -3,62 +3,84 @@ package standalone
...
-		if err := db.Close(); err != nil {
-			errs = multierr.Append(errs, err)
-		}
+	err := s.Container.Close()
+
+	for _, c := range s.controllers {
+		err = multierr.Append(err, c.Close())
 	}
-	return errs
+	return err
```
nahguam, line 87: could do `Combine` like in the other places

andrasbeni: Do you suggest accumulating the errors in an array and then calling Combine after the loop?

merlimat: @nahguam The `Append()` is really the same as `Combine()` and it's in the best practice when adding from a loop: 

https://github.com/uber-go/multierr/blob/445999037311d4b33da9ece1730660e666862821/error.go#L46-L54

PR #78, opened 2022-11-29
Add context labels to go routines

merlimat:
To help when getting a stack trace or heap dump, we should identify the go routine with context

### DIFF
--- a/common/pprof.go
+++ b/common/pprof.go
@@ -10,8 +10,7 @@ import (
 func DoWithLabels(labels map[string]string, f func()) {
 	var l []string
 	for k, v := range labels {
-		l = append(l, k)
-		l = append(l, v)
+		l = append(l, k, v)
 	}
 
 	pprof.Do(


oxia/internal/shard_manager.go
```diff
@@ -49,12 +50,19 @@ func (s *shardManagerImpl) Start() {
...
-	go func() {
-		if _, ok := <-s.closeC; !ok {
-			cancel()
-		}
-	}()
+	go pprof.Do(context.Background(),
+		pprof.Labels("oxia", "receive-shard-updates"),
+		func(_ context.Context) {
+			s.receiveWithRecovery(ctx, readyC)
+		})
```
nahguam: There's lots of repetitive stanzas like this, should we add a util function for this?

```go
func pprofDo(f func(), labels ...string) {
	pprof.Do(context.Background(),
		pprof.Labels(labels...),
		func(_ context.Context) {
			f()
		})
}
```

merlimat: Very good point. I'll add one

merlimat: Added `DoWithLabels()`

common/pprof.go
```diff
@@ -0,0 +1,23 @@
...
+	"runtime/pprof"
+)
+
+// DoWithLabels attaches the labels to the current go-routine Pprof context,
+// for the duration of the call to f
+func DoWithLabels(labels map[string]string, f func()) {
+	var l []string
+	for k, v := range labels {
+		l = append(l, k)
+		l = append(l, v)
```
nahguam: nit

```suggestion
		l = append(l, k, v)
```

PR #79, opened 2022-11-29
Added CLI option to enable pprof in every component

merlimat:
This is stacked on top of #78 

Added the `--profile` flag to enable Go pprof, so that we can get stack traces, heap dump, profiler and flame graphs


server/main.go
```diff
@@ -34,5 +34,10 @@ func main(cmd *cobra.Command, args []string) {
 			Msg("Failed to start the server")
 	}
 
-	common.WaitUntilSignal(server)
+	profiler := common.RunProfiling()
```
nahguam, line 36: Should these only activate if the user provided the `profile` flag?

merlimat: Inside the `RunProfiling()` is checking the global flag set in the CLI

nahguam: ah so it is! 👍

PR #83, opened 2022-11-30
Add docker entrypoint

nahguam:
With entrypoint:

```
docker run oxia:latest standalone
```

```
spec:
  containers:
  - image: oxia:latest
    args:
    - server
    - ...
```

Without entrypoint:

```
docker run oxia:latest oxia standalone
```

```
spec:
  containers:
  - image: oxia:latest
    command:
    - oxia
    - server
    - ...
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,6 @@ RUN apk add build-base
 RUN cd /oxia-src \
     && make
 
-
-
 FROM alpine:3.16.1
 
 RUN mkdir /oxia
@@ -24,3 +22,4 @@ WORKDIR /oxia
 COPY --from=build /oxia-src/bin/oxia /oxia/bin/oxia
 ENV PATH=$PATH:/oxia/bin
 
+ENTRYPOINT ["oxia"]


Dockerfile
```diff
@@ -24,3 +22,4 @@ WORKDIR /oxia
 COPY --from=build /oxia-src/bin/oxia /oxia/bin/oxia
 ENV PATH=$PATH:/oxia/bin
 
+ENTRYPOINT ["oxia"]
```
merlimat, line 25: Wouldn't this make more difficult to spin up a container to use a client interactively?

eg:

```
$ docker run -it oxia:latest

$ oxia client put -k a -v 1
$ oxia client put -k b -v 2
```

nahguam: Yes, that's true. I was only considering non-interactive and fewer args in the container spec. I'll close this.

Side note - the client won't work outside the cluster https://www.notion.so/streamnativeio/Notes-82d5da6b207a44fe89b25f6042e41283

You'd need to create a pod:

```
kubectl run oxia --image oxia:latest -it

/oxia # oxia client put -k a -v 1
```

PR #90, opened 2022-12-01
Coordinator

merlimat:
The first iteration of the coordinator component. 

Includes:
 * Struct definitions
 * Interface for metadata provider (eg: we'll need an implementation that writes back using the K8S APIs)
 * Initial leader election
 * Sending shard-assignments updates to storage nodes

Missing
 * Any tests
 * Health-check
 * Any follow-up on leader crashing

### DIFF
--- a/coordinator/node_controller.go
+++ b/coordinator/node_controller.go
@@ -22,6 +22,8 @@ type nodeController struct {
 	clientPool               common.ClientPool
 	log                      zerolog.Logger
 	closed                   atomic.Bool
+	ctx                      context.Context
+	cancel                   context.CancelFunc
 }
 
 func NewNodeController(addr string, shardAssignmentsProvider ShardAssignmentsProvider, clientPool common.ClientPool) NodeController {
@@ -34,6 +36,8 @@ func NewNodeController(addr string, shardAssignmentsProvider ShardAssignmentsPro
 			Str("addr", addr).
 			Logger(),
 	}
+
+	nc.ctx, nc.cancel = context.WithCancel(context.Background())
 
 	go common.DoWithLabels(map[string]string{
 		"oxia": "node-controller",
@@ -74,7 +78,7 @@ func (n *nodeController) sendAssignmentsUpdates() error {
 		return err
 	}
 
-	stream, err := rpc.ShardAssignment(context.Background())
+	stream, err := rpc.ShardAssignment(n.ctx)
 	if err != nil {
 		return err
 	}
@@ -103,5 +107,6 @@ func (n *nodeController) sendAssignmentsUpdates() error {
 
 func (n *nodeController) Close() error {
 	n.closed.Store(true)
+	n.cancel()
 	return nil
 }


coordinator/shard_status.go
```diff
@@ -0,0 +1,49 @@
...
+
+// MarshalJSON marshals the enum as a quoted json string
+func (s ShardStatus) MarshalJSON() ([]byte, error) {
+	buffer := bytes.NewBufferString(`"`)
+	buffer.WriteString(toString[s])
+	buffer.WriteString(`"`)
+	return buffer.Bytes(), nil
+}
+
+// UnmarshalJSON unmashals a quoted json string to the enum value
```
andrasbeni:
```suggestion
// UnmarshalJSON unmarshals a quoted json string to the enum value
```

coordinator/shard_status.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type ShardStatus uint16
+
+const (
+	ShardStatusUnknown = iota
```
andrasbeni:
```suggestion
	ShardStatusUnknown ShardStatus = iota
```

otherwise the test
```
func TestUnknown(t *testing.T) {
	var s ShardStatus
	err := s.UnmarshalJSON([]byte("\"Unknown2\""))
	if err != nil {
		assert.NoError(t, err)
	}
	assert.Equal(t, ShardStatusUnknown, s)
}
``` 
will fail with 
```
Expected :int(0)
Actual   :server.ShardStatus(0x0)
```

server/public_rpc_server.go
```diff
@@ -38,27 +41,66 @@ func NewPublicRpcServer(port int, shardsDirector ShardsDirector, assignmentDispa
 }
 
 func (s *PublicRpcServer) ShardAssignments(_ *proto.ShardAssignmentsRequest, srv proto.OxiaClient_ShardAssignmentsServer) error {
-	return s.assignmentDispatcher.AddClient(srv)
+	s.log.Debug().
```
andrasbeni, line 44: Not strictly coordinator code, so could have been in another PR.

coordinator/node_controller.go
```diff
@@ -0,0 +1,109 @@
...
+type NodeController interface {
+	io.Closer
+}
+
+type nodeController struct {
+	addr                     string
+	shardAssignmentsProvider ShardAssignmentsProvider
+	clientPool               common.ClientPool
+	log                      zerolog.Logger
+	closed                   bool
```
andrasbeni: We'll need a mutex to guard `closed`.

coordinator/shard_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	return res.HeadIndex, nil
+}
+
+func (s *shardController) selectNewLeader(fenceResponses map[ServerAddress]*proto.EntryId) (
+	leader ServerAddress, followers map[string]*proto.EntryId) {
+	// Select all the nodes that have the highest entry in the wal
+	var currentMax int64 = -1
+	var candidates []ServerAddress
+
+	for a, e := range fenceResponses {
```
andrasbeni: Can you please use more descriptive names like `address` and `headIndex`?

coordinator/coordinator.go
```diff
@@ -0,0 +1,231 @@
...
+
+	for pb.Equal(currentValue, c.assignments) {
+		// Wait on the condition until the assignments get changed
+		c.assignmentsChanged.Wait()
+	}
+
+	return c.assignments
+}
+
+func (c *coordinator) InitiateLeaderElection(shard uint32, newEpoch int64) error {
```
andrasbeni, line 162: Would it make sense to push this logic to the MeatadataProvider and call it directly from shardProvider?

server/public_rpc_server.go
andrasbeni: I don't insist on removing this from the PR, but maybe next time you can consider breaking down the pull request to multiple, smaller PRs.

coordinator/coordinator.go
merlimat: The `MetadataProvider` is meant to abstract the storage of the cluster status. We'll probably have multiple implementations (eg: K8S API server, bare Etcd, ZK or even possibly a custom Raft group). 

There shouldn't be any other logic in that interface.

PR #91, opened 2022-12-02
Move all cobra commands under cmd

nahguam:
There were multiple `main.go` files and `main` functions.
The files in the `cmd` directory were using the `main` package.
Now all cobra commands are in the `cmd` package.

### DIFF
--- a/common/logger.go
+++ b/common/logger.go
@@ -1,9 +1,12 @@
 package common
 
 import (
+	"encoding/json"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/rs/zerolog/pkgerrors"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "oxia/cmd"
+
+func main() {
```
merlimat: Couldn't this reside in the `cmd` package? Maybe renamed into `main` ?

nodece: Good catch! We only need to handle the `cmd/root.go`:

1. Rename `cmd/root.go` to `cmd/main.go`
2. Rename the package name from `cmd` to `main`
3. Rename the `Execute()` to `main()`

nahguam: Ok, I've put main back in cmd

PR #96, opened 2022-12-04
Better handling of closing stream

merlimat:
In the `FollowerCursor`, we should keep a `CancelFunc` to close the stream when we want to close the cursor. Otherwise we would have the goroutine stuck on the `stream.Recv()` operation.

~Note: this is stacked on top of #97 to use the `Backoff` methods~

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module oxia
 go 1.19
 
 require (
+	github.com/cenkalti/backoff/v4 v4.2.0
 	github.com/cockroachdb/pebble v0.0.0-20221104214247-8dc60b62ebbf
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 require (
 	github.com/DataDog/zstd v1.4.5 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/cenkalti/backoff/v4 v4.2.0 // indirect
```
nahguam: I think it just needs a `go mod tidy`

merlimat: Done

PR #99, opened 2022-12-05
Create controller service

nahguam:


cmd/flag/flag.go
```diff
@@ -0,0 +1,18 @@
+package flag
+
+import (
+	"github.com/spf13/cobra"
+	"oxia/operator/resource"
+)
```
teabot, line 6: Can/should we later move other common flags here? I'm thinking of `service-address`

cmd/flag/flag.go
```diff
@@ -0,0 +1,18 @@
...
+
+func PublicPort(cmd *cobra.Command, conf *int) {
+	cmd.Flags().IntVarP(conf, "public-port", "p", resource.PublicPort(), "Public service port")
```
teabot, line 9: If you have sub-commands you may wish to use `PersistentFlags`:
```suggestion
	cmd.PersistentFlags().IntVarP(conf, "public-port", "p", resource.PublicPort(), "Public service port")
```

operator/resource/config.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+var Ports = map[string]int{
+	PublicPortName:   6648,
+	InternalPortName: 6649,
+	MetricsPortName:  8080,
+}
+
+func PublicPort() int {
+	return Ports[PublicPortName]
```
merlimat, line 16: If we're exposing through these, we should maybe rename to `ports`

nahguam: They're being used in a later PR

PR #102, opened 2022-12-05
Fork wal library

andrasbeni:
Motivation:
https://github.com/streamnative/oxia/pull/74#issue-1467266070

### DIFF
--- a/server/wal/log.go
+++ b/server/wal/log.go
@@ -1,6 +1,28 @@
-package wal
+/*
+The MIT License (MIT)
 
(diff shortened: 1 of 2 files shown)


server/wal/log.go
```diff
@@ -0,0 +1,902 @@
+package wal
+
+// Based on https://github.com/tidwall/wal
```
merlimat: We should include the original license and copyright header: https://github.com/tidwall/wal/blob/master/LICENSE

PR #104, opened 2022-12-06
Commands for [un]installing CRD and controller

nahguam:


cmd/main.go
```diff
@@ -28,6 +29,7 @@ func init() {
 
 	rootCmd.AddCommand(client.Cmd)
 	rootCmd.AddCommand(controller.Cmd)
+	rootCmd.AddCommand(operator.Cmd)
```
merlimat, line 32: Can we remove `controller` from top level command?

nahguam: `oxia controller` is to start the controller process (equivalent of `oxia server`, `oxia standalone`). Why do you want to remove it? 

I grouped the operator administrative tasks under `oxia operator`, so we have: `oxia operator crd [un]install` & `oxia operator controller [un]install`

PR #105, opened 2022-12-06
[Coordinator] Implemented health check to storage nodes and triggering leader election

merlimat:
Added the logic in the `NodeController` to perform health-checks for the storage nodes and trigger a new election when the leader of a given shard has failed.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"oxia/cmd/client"
 	"oxia/cmd/controller"
+	"oxia/cmd/operator"
 	"oxia/cmd/server"
 	"oxia/cmd/standalone"
 	"oxia/common"
@@ -28,6 +29,7 @@ func init() {
 
 	rootCmd.AddCommand(client.Cmd)
 	rootCmd.AddCommand(controller.Cmd)
+	rootCmd.AddCommand(operator.Cmd)
 	rootCmd.AddCommand(server.Cmd)
 	rootCmd.AddCommand(standalone.Cmd)
 }
--- a/cmd/operator/cmd.go
+++ b/cmd/operator/cmd.go
@@ -1,0 +1,20 @@
+package operator
+
+import (
+	"github.com/spf13/cobra"
+	"oxia/cmd/controller"
+	"oxia/cmd/operator/crd"
+)
+
+var (
+	Cmd = &cobra.Command{
+		Use:   "operator",
+		Short: "Install or Uninstall CRD and controller",
+		Long:  `Install or Uninstall CRD and controller`,
+	}
+)
+
+func init() {
+	Cmd.AddCommand(crd.Cmd)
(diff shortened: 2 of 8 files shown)


coordinator/node_controller.go
```diff
@@ -17,63 +34,150 @@ type NodeController interface {
...
 	}, func() {
-		nc.sendAssignmentsUpdatesWithRetries()
+		ticker := time.NewTicker(healthCheckProbeInterval)
+
+		for {
+			select {
+			case <-ticker.C:
+				pingCtx, pingCancel := context.WithTimeout(ctx, healthCheckProbeTimeout)
+
+				res, err := health.Check(pingCtx, &grpc_health_v1.HealthCheckRequest{Service: ""})
```
nahguam, line 119: We'll also be using this as the liveness and readiness probe for the pods. If they fail more than the readiness probe policy, it'll be taken out of the service (i.e. won't be available at `{service name}-{statefulset ordinal}`) until it recovers. If they fail more than the liveness probe policy then the pod will be restarted.

This will be competing with this check. We might want to use different service names to the health check for the different levels of degradation that would trigger the respective responses by kube/oxia coordinator - liveness, readiness & leader election.

merlimat: >  it'll be taken out of the service (i.e. won't be available at {service name}-{statefulset ordinal}) 

If it fails the readiness probe, it will be taken out of the service rotation, though the individual pod will still respond. 
The effect will be that new client requests for service discovery will not get sent to this pod. 

> This will be competing with this check. 

I agree that at some point we'd want to have more fine-grained checks, though I don't think there is any competition here.

The 2 checks are used for different purposes and do not affect each other. In this case, we just need to trigger a leader election when a leader fails.

nahguam: If the healthcheck request is sent to the k8s DNS name then the pod won't respond as the k8s network won't know where to route the request to, as a restarting pod or one failing its readiness probe will not be present in the service. I've seen approaches in the past to cache the IP address after the first resolution.

By 'competing' I meant that it's just as likely that the healthcheck will timeout rather return a NOTSERVING response. I guess I was thinking that maybe failing to resolve the DNS name might be an alternative approach to triggering leader election.

Anyway, I think what's here is fine to start

merlimat: > By 'competing' I meant that it's just as likely that the healthcheck will timeout rather return a NOTSERVING response. I guess I was thinking that maybe failing to resolve the DNS name might be an alternative approach to triggering leader election.

That's ok if we're the failing the healtch-check when the pod is not ready. When I'm saying we're not competing, I mean we're not using the outcome of the check to do anything related to this pod, rather we want to make a decision about the shards that are running on the pod. 

If it's responding NOT_SERVING or just is not reachable, it really won't matter to us, as we need to trigger a leader election for the shard in both cases.

PR #107, opened 2022-12-07
CRD install implementation

nahguam:


examples/resources/crd.yaml
```diff
@@ -42,3 +76,21 @@ spec:
   serverReplicas: 1
   shardCount: 1
   replicationFactor: 1
+status:
+  shards:
+    - id: 0
```
nahguam, line 81: In the k8s status I've modelled it as an array. I think it's more idiomatic that way.

Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 ENV PATH=$PATH:/go/bin
 ADD . /oxia-src
 
-RUN apk add build-base
+RUN apk add build-base bash
```
merlimat: We should add both to the `apk add` in line 4

examples/resources/crd.yaml
merlimat: 👍  Makes sense

PR #116, opened 2022-12-10
Perform fencing in parallel

merlimat:
(this is stacked on top of #114)

When fencing the quorum, we should do it in parallel so that 1 single unresponsive node will not cause the leader election to be blocked until a timeout occurs. 

Once we have reached a majority of the responses from the ensemble, we wait another short time (eg: 100ms) to ensure all healthy servers are added to the initial set.

All the servers that don't respond to the fence will be retried later and added to the leader once they are available. This logic is not there yet in the coordinator.

### DIFF
--- a/coordinator/impl/mock_test.go
+++ b/coordinator/impl/mock_test.go
@@ -201,14 +201,16 @@ func (r *mockRpcProvider) GetShardAssignmentStream(ctx context.Context, node Ser
 
 func (r *mockRpcProvider) Fence(ctx context.Context, node ServerAddress, req *proto.FenceRequest) (*proto.FenceResponse, error) {
 	r.Lock()
-	defer r.Unlock()
 
 	s := r.getNode(node)
 	s.fenceRequests <- req
 
 	if s.err != nil {
+		r.Unlock()
 		return nil, s.err
 	}
+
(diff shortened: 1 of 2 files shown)


coordinator/impl/shard_controller.go
```diff
@@ -144,40 +151,96 @@ func (s *shardController) electLeader() error {
...
-			Interface("server-address", sa).
-			Interface("entry-id", entryId).
-			Int64("epoch", s.shardMetadata.Epoch).
-			Msg("Processed fence response")
+	res := make(map[ServerAddress]*proto.EntryId)
+	var err error
+
+	// Wait for a majority to respond
+	for successResponses < majority && totalResponses < ensembleSize {
+		r := <-ch
```
andrasbeni, line 207: Since we never close the channel, this could block forever.
But I'm not sure if it really is a problem because if less than the majority responds at all, the shard's ensemble has bigger problems than a blocked vthread.

merlimat: This only waits for 1 round of fence requests. The requests can be:
 * Cancelled centrally (eg: the shard-controller is closed)
 * Timed-out -- Every requests has a 30sec timeout

PR #118, opened 2022-12-12
Coordinator Service

nahguam:
The beginnings of the coordinator service - just using memory metadata provider and assuming the service address match up to the statefulset service.

To keep things consistent with the other services and to avoid name conflicts, I've moved everything that was in the `coordinator` package into `coordinator/impl`. I realise this is a terrible name. Very happy to hear suggestions.

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -14,7 +14,7 @@ type Config struct {
 	MetricsPort         int
 	Name                string
 	ReplicationFactor   uint32
-	ShardsCount         uint32
+	ShardCount          uint32
 	ServerReplicas      uint32
 }
 
@@ -24,7 +24,7 @@ func NewConfig() Config {
 		MetricsPort:         resource.MetricsPort.Port,
 		Name:                "oxia",
 		ReplicationFactor:   3,
-		ShardsCount:         3,
+		ShardCount:          3,
 		ServerReplicas:      3,
 	}
 }
@@ -47,7 +47,7 @@ func New(config Config) (*Coordinator, error) {
 	// perhaps a controller -> coordinator RPC
 	clusterConfig := impl.ClusterConfig{
 		ReplicationFactor: config.ReplicationFactor,
-		ShardsCount:       config.ShardsCount,
+		ShardCount:        config.ShardCount,
 		StorageServers:    storageServers(config),
 	}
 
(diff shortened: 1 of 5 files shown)


coordinator/coordinator.go
```diff
@@ -1,246 +1,90 @@
...
-		ClusterConfig:    clusterConfig,
-		shardControllers: make(map[uint32]ShardController),
-		nodeControllers:  make(map[string]NodeController),
-		clientPool:       common.NewClientPool(),
-		log: log.With().
-			Str("component", "coordinator").
-			Logger(),
+	metadataProvider := impl.NewMetadataProviderMemory()
+	// TODO Eventually this will need a more dynamic method of updating the config when it changes
+	// perhaps a controller -> coordinator RPC
```
merlimat, line 47: We could take a config file with the cluster config and, in K8S env, have the config file stored in a config map.

The config map will keep the file updated in the coordinator pod.

coordinator/coordinator.go
```diff
@@ -1,246 +1,90 @@
...
+	ServerReplicas      uint32
 }
 
-type NodeAvailabilityListener interface {
-	NodeBecameUnavailable(node ServerAddress)
+func NewConfig() Config {
+	return Config{
+		InternalServicePort: resource.InternalPort.Port,
+		MetricsPort:         resource.MetricsPort.Port,
+		Name:                "oxia",
```
merlimat, line 25: What is the `name` referring to here?

coordinator/coordinator.go
```diff
@@ -1,246 +1,90 @@
...
+		ReplicationFactor:   3,
+		ShardCount:          3,
+		ServerReplicas:      3,
```
merlimat, line 28: Taking ServerReplicas here implies assuming K8S statefulset name style. We could instead take a list of `ServerAddress` and probably take it from a config file.

This way we could use the same `oxia coordinator` outside k8s (barring configuration on how to access metadata provider).

coordinator/coordinator.go
nahguam: Yes, I've used that approach in the past :thumbsup: you can mount the cm data as a file in the pod.

coordinator/coordinator.go
nahguam: This is what the names of the other resources will be - servers statefulset, Coordinator deployment, services etc. The host names will be derived from it.

coordinator/coordinator.go
nahguam: Yes, I was just using this to get something working. I like the file idea. I'll do that tomorrow in a follow up if that's OK?

merlimat: 👍

PR #119, opened 2022-12-12
 Allow full range of log levels

teabot:
# Motivation
Sometimes info and debug is not enough.

# Changes
* Removed `-d` flag from CLI
* Added `-l` flag to permit usage of any one of the [rs/zerolog](https://github.com/rs/zerolog) levels
* Defaults to `info` as before
* Added test cases

### DIFF
--- a/cmd/main_test.go
+++ b/cmd/main_test.go
@@ -17,7 +17,7 @@ func TestCall_LogLevel_Default(t *testing.T) {
 	}
 	err := rootCmd.Execute()
 	assert.Equal(t, common.DefaultLogLevel, captured)
-	assert.ErrorIs(t, err, nil)
+	assert.NoError(t, err)
 }
 
 func TestCall_LogLevel(t *testing.T) {


cmd/main.go
```diff
@@ -14,15 +15,23 @@ import (
...
 
+type LogLevelError string
+
+func (l LogLevelError) Error() string {
+	return fmt.Sprintf("unknown log level (%s)", string(l))
+}
+
 func init() {
-	rootCmd.PersistentFlags().BoolVarP(&common.LogDebug, "log-debug", "d", false, "Enable debug logs")
+	rootCmd.PersistentFlags().StringVarP(&logLevelStr, "log-level", "g", common.DefaultLogLevel.String(), "Set logging level [disabled|trace|debug|info|warn|error|fatal|panic]")
```
andrasbeni: Why `g`? `-l` would be less surprising in my opinion.

cmd/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"oxia/common"
+	"testing"
+)
+
+func TestCall_LogLevel_Default(t *testing.T) {
+	t.Run("default-log-level", func(t *testing.T) {
```
andrasbeni: What's the point of wrapping the whole test in a closure? It would work just the same without it, wouldn't it.

common/logger.go
```diff
@@ -11,9 +11,11 @@ import (
 	"time"
 )
 
+const DefaultLogLevel zerolog.Level = zerolog.InfoLevel
```
andrasbeni:
```suggestion
const DefaultLogLevel = zerolog.InfoLevel
```

cmd/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+				assert.Equal(t, test.expectedLevel, common.LogLevel)
+				return nil
+			}
+			err := rootCmd.Execute()
+			if err == nil {
+				assert.True(t, invoked)
+			} else {
+				assert.False(t, invoked)
+			}
+			assert.ErrorIs(t, test.expectedErr, err)
```
andrasbeni: Actually, `ErrorIs` checks if its third parameter is in the chain of the second parameter. This is not a problem when the two are equal, but it's better to get used to the intended usage. 
```suggestion
			assert.ErrorIs(t, err, test.expectedErr)
```

cmd/main.go
teabot: Yep, not sure why I didn't go to `l` — changing.

cmd/main_test.go
teabot: copy pasta error — fixing

cmd/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+func TestCall_LogLevel_Default(t *testing.T) {
+	var captured zerolog.Level
+	rootCmd.SetArgs([]string{})
+	rootCmd.RunE = func(cmd *cobra.Command, args []string) error {
+		captured = common.LogLevel
+		return nil
+	}
+	err := rootCmd.Execute()
+	assert.Equal(t, common.DefaultLogLevel, captured)
+	assert.ErrorIs(t, err, nil)
```
nahguam: nit

```suggestion
	assert.NoError(t, err)
```

cmd/main.go
```diff
@@ -14,15 +15,23 @@ import (
...
-	rootCmd = &cobra.Command{
-		Use:   "oxia",
-		Short: "Short description",
-		Long:  `Long description`,
+	logLevelStr string
+	rootCmd     = &cobra.Command{
+		Use:     "oxia",
+		Short:   "Short description",
+		Long:    `Long description`,
+		PreRunE: configureLogLevel,
```
merlimat, line 23: We could also call the `common.ConfigureLogger()` from this hook

common/logger.go
```diff
@@ -11,9 +11,11 @@ import (
 	"time"
 )
 
+const DefaultLogLevel = zerolog.InfoLevel
+
 var (
-	// LogDebug Used for flags
-	LogDebug bool
+	// LogLevel Used for flags
+	LogLevel zerolog.Level
```
merlimat, line 18: (for a different PR) Should we also improve the `LogJson` argument to make it a kind of `--log-format json|text|....` ?

PR #120, opened 2022-12-12
Session Manager

andrasbeni:
https://www.notion.so/streamnativeio/Design-for-session-management-b3be2beca3a941bf8c2313a0ddeaf8b3

### DIFF
--- a/server/session.go
+++ b/server/session.go
@@ -34,10 +34,10 @@ func startSession(sessionId SessionId, sessionMetadata *proto.SessionMetadata, s
 		closeCh:     make(chan error),
 		log: sm.log.With().
 			Str("component", "session").
-			Str("session-id", hexId(sessionId)).Logger(),
+			Int64("session-id", int64(sessionId)).Logger(),
 	}
 	go s.waitForHeartbeats()
-	s.log.Info().Msg("Session started")
+	s.log.Debug().Msg("Session started")
 	return s
 }
 
--- a/server/session_manager.go
+++ b/server/session_manager.go
@@ -72,6 +72,7 @@ func NewSessionManager(shardId uint32, controller *leaderController) SessionMana
 		log: log.With().
 			Str("component", "session-manager").
 			Uint32("shard", shardId).
+			Int64("epoch", controller.epoch).
 			Logger(),
 	}
 }
@@ -211,7 +212,7 @@ func (sm *sessionManager) readSessions() (map[SessionId]*proto.SessionMetadata, 
 		if metaEntry.Status != proto.Status_OK {
 			sm.log.Warn().
 				Str("key", key).
-				Str("status", fmt.Sprintf("%#v", metaEntry.Status)).
+				Stringer("status", metaEntry.Status).
 				Msgf("error reading session metadata")
 			continue
 		}
@@ -229,7 +230,7 @@ func (sm *sessionManager) readSessions() (map[SessionId]*proto.SessionMetadata, 
 		if err != nil {
 			sm.log.Warn().
 				Err(err).
-				Int32("session-id", int32(sessionId)).
+				Int64("session-id", int64(sessionId)).
 				Str("key", key).
 				Msgf("error unmarshalling session metadata")
 			continue


proto/client.proto
```diff
@@ -288,4 +306,27 @@ enum Status {
...
+message CreateSessionRequest {
+  uint32 shard_id = 1;
+  uint32 session_timeout_ms = 2;
 }
+
+message CreateSessionResponse {
+  uint32 session_id = 1;
+}
+
+message Heartbeat {
```
merlimat: Maybe name it `SessionHeartBeat`?

proto/client.proto
```diff
@@ -288,4 +306,27 @@ enum Status {
...
+  uint32 session_id = 1;
```
merlimat: We should probably also include the `shard_id` since the session id would not be unique across shards

common/condition_test.go
```diff
@@ -92,6 +92,7 @@ func TestCondSignal(t *testing.T) {
 }
 
 func TestCondSignalGenerations(t *testing.T) {
+	t.SkipNow()
```
merlimat: This we should remove

server/leader_controller.go
```diff
@@ -175,6 +178,11 @@ func (lc *leaderController) Fence(req *proto.FenceRequest) (*proto.FenceResponse
 		return nil, err
 	}
 
+	err = lc.sessionManager.CloseShard(lc.shardId)
```
merlimat: Instead of 1 session manager that handles all the shards, it's simpler to have a per-shard session-manager that is initialized & closed in the leader-controller

server/leader_controller.go
```diff
@@ -424,6 +441,10 @@ func (lc *leaderController) Read(request *proto.ReadRequest) (*proto.ReadRespons
 		}
 	}
 
+	return lc.readWithoutLocking(request)
```
merlimat: remove the `{}` for the lock&unlock

server/public_rpc_server.go
```diff
@@ -125,6 +125,47 @@ func (s *PublicRpcServer) GetNotifications(req *proto.NotificationsRequest, stre
 	return err
 }
 
+func (s *PublicRpcServer) CreateSession(ctx context.Context, req *proto.CreateSessionRequest) (*proto.CreateSessionResponse, error) {
+	s.log.Debug().
+		Str("peer", common.GetPeer(ctx)).
+		Interface("req", req).
+		Msg("Create session request")
+	res, err := s.shardsDirector.GetSessionManager().CreateSession(req)
```
merlimat: It would be better to always go through the leader-controller

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+
+var ErrorInvalidSessionTimeout = errors.New("invalid session timeout")
+
+type SessionId uint64
+
+func SessionKey(sessionId SessionId) string {
+	return KeyPrefix + strconv.FormatUint(uint64(sessionId), 10) + "/"
+}
+
+func KeyToId(key string) (SessionId, error) {
```
merlimat, line 34: We could use the same hex notation as for https://github.com/streamnative/oxia/blob/main/server/kv/notifications_tracker.go#L57

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+		return nil, errors.Wrap(ErrorInvalidSessionTimeout, fmt.Sprintf("timeoutMs=%d", request.SessionTimeoutMs))
+	}
+	sm.Lock()
+	defer sm.Unlock()
+	if _, initialized := sm.sessionCounters[request.ShardId]; !initialized {
+		return nil, errors.Errorf("session management for shard %d not initialized", request.ShardId)
+	}
+	sm.sessionCounters[request.ShardId]++
+
+	sessionId := (SessionId(request.ShardId) << 32) + sm.sessionCounters[request.ShardId]
```
merlimat: There is no need to have a unique session id across shards, though here we're restricting the counter to effectively be 32 bits, within a single shard. 

As I mentioned on the design doc, the session creation must be recorded in the WAL. At that point the Id for the session could be the offset where the session creation happens is recorded.

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+	return proto.Status_OK, nil
+
+}
+
+type LeaderControllerSupplier func(shardId uint32) (LeaderController, error)
+
+// --- SessionManager
+
+type SessionManager interface {
```
merlimat, line 49: As I mentioned above, I think we need to have 1 manager/controller per shard.

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+	} else {
+		s.attached = true
+		s.closeCh = closeChannel
+	}
+
+	s.Unlock()
+	sm.Unlock()
+
+	for {
+		heartbeat, err := stream.Recv()
```
merlimat: Doing the `Recv()` from the goroutine that is handling the stream will block the closing of the session manager (and thus the overall shutdown of the process). 

We need to handle it like: https://github.com/streamnative/oxia/blob/main/server/follower_controller.go#L271

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+	shardId     uint32
+	timeout     time.Duration
+	sm          *sessionManager
+	attached    bool
+	heartbeatCh chan *proto.Heartbeat
+	closeCh     chan error
+	log         zerolog.Logger
+}
+
+func (s *session) close(delete bool) error {
```
merlimat: When would be a session closed but not deleted? The 2 operations should probably be split.

server/kv/db.go
```diff
@@ -23,10 +23,15 @@ const (
...
+type PutCustomizer interface {
+	CheckApplicability(WriteBatch, *proto.PutRequest) (proto.Status, error)
+	AdditionalData(*proto.PutRequest) *proto.PutRequest
+}
+
 type DB interface {
 	io.Closer
 
-	ProcessWrite(b *proto.WriteRequest, commitIndex int64, timestamp uint64) (*proto.WriteResponse, error)
+	ProcessWrite(b *proto.WriteRequest, commitIndex int64, timestamp uint64, putCustomizer PutCustomizer) (*proto.WriteResponse, error)
```
merlimat: Instead of a `PutCustomizer`, couldn't we pass in a callback that gets called with the request and the writeBatch?

Also, it won't be also needed on put but on deletes too, to cleanup the session-shadow node.

server/kv/db.go
```diff
@@ -253,6 +258,25 @@ func (d *db) applyPut(batch WriteBatch, notifications *notifications, putReq *pr
 	} else if err != nil {
 		return nil, errors.Wrap(err, "oxia db: failed to apply batch")
 	} else {
+		if putCustomizer != nil {
+			status, err := putCustomizer.CheckApplicability(batch, putReq)
+			if err != nil {
+				return nil, err
```
merlimat: In case of error, we shouldn't be failing the entire batch, just the specify putRequest

server/session_manager.go
```diff
@@ -0,0 +1,460 @@
...
+	if s.heartbeatCh != nil {
+		close(s.heartbeatCh)
+		s.heartbeatCh = nil
+	}
+
+	if !delete {
+		return nil
+	}
+	// Delete ephemeral data associated with this session
+	controller, err := s.sm.controllerSupplier(s.shardId)
```
merlimat: The deleting of a session should be written in the WAL.

proto/client.proto
```diff
@@ -290,8 +308,31 @@ enum Status {
...
+}
+
+message Heartbeat {
+  uint64 session_id = 2;
+}
+
+message KeepAliveResponse {}
+
+message CloseSessionRequest {
+  uint64 session_id = 1;
```
merlimat: Should have `shard_id` too

proto/client.proto
nahguam: `KeepAliveRequest` to be consistent with the naming of the other rpcs

common/condition_test.go
andrasbeni: Yes, but that test constantly goes into a deadlock. We need to fix the condition object, but I was not able to figure out what actually is wrong.

andrasbeni:
```
goroutine 29 [chan receive]:
oxia/common.TestCondSignalGenerations(0xc000338ea0)
        /Users/andrasbeni/oxia/common/condition_test.go:109 +0x3a4
testing.tRunner(0xc000338ea0, 0x100948bf0)
        /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1446 +0x18c
created by testing.(*T).Run
        /opt/homebrew/Cellar/go/1.19.3/libexec/src/testing/testing.go:1493 +0x560

goroutine 40 [select]:
oxia/common.(*conditionContext).Wait(0xc00035c930, {0x100950030, 0xc0001a6018})
        /Users/andrasbeni/oxia/common/condition.go:71 +0x138
oxia/common.TestCondSignalGenerations.func1(0x6)
        /Users/andrasbeni/oxia/common/condition_test.go:104 +0xa8
created by oxia/common.TestCondSignalGenerations
        /Users/andrasbeni/oxia/common/condition_test.go:101 +0x388

goroutine 33 [select]:
oxia/common.(*conditionContext).Wait(0xc00035c930, {0x100950030, 0xc0001a6018})
        /Users/andrasbeni/oxia/common/condition.go:71 +0x138
oxia/common.TestCondSignalGenerations.func1(0x5)
        /Users/andrasbeni/oxia/common/condition_test.go:104 +0xa8
created by oxia/common.TestCondSignalGenerations
        /Users/andrasbeni/oxia/common/condition_test.go:101 +0x388
```

server/session_manager.go
andrasbeni: On line 144 below, the session creation is recorded in the WAL. 
How do you think using the offset as a session ID would work? In this current design, the session is recorded as the key of a Put operation. I can only imagine that I can figure out what the offset will be before creating the entry if I do all this within WAL. And I think WAL should not be tied to session management in this way.

How about keeping the counter (persisting its state whenever a new session is created) but using uint64 and always using it together with shard ID? That would still allow way more sessions than the 4 billion allowed by uint32.

server/session_manager.go
andrasbeni: When the server shuts down.
I've split them into `closeChannels` which will be used every time we stop listening for updates and `delete` which will only be used when we do this because the client requested to close the session or timed out.

server/session_manager.go
andrasbeni: Yes, that's why I call this method in a separate go routine (`go sm.receiveHeartbeats(server, closeChannel)`) on line 171.

server/public_rpc_server.go
andrasbeni: I agree. The tradeoff here was that KeepAlive has no initial request that would contain the shard ID, so we need to have a "global" object to handle it outside the context of a shard. But as it looks to me, you'd want to pass the shard ID in the context like it is done in the case of AddEntries, right?

server/session_manager.go
andrasbeni: It is. See line 384

server/leader_controller.go
andrasbeni: https://github.com/streamnative/oxia/pull/180

server/kv/db.go
andrasbeni: When `batch.Put(...)` returns an error below (line 299), then it fails the entire batch. Why would we treat it differently when a `batch.Get(...)` returns an error (other than `ErrorKeyNotFound`, which is treated differently)?

server/public_rpc_server.go
merlimat: Yes, it's a bummer that there's no better way of doing that with gRPC, though it's probably the best option.

server/session_manager.go
merlimat: > How about keeping the counter (persisting its state whenever a new session is created) but using uint64 and always using it together with shard ID?

Though that would still mean to go through the WAL to increment the id. 

I think it would be perfectly fine to pass the "current offset" for a WriteBatch to the DB. 

The only thing we need to make sure is to have the `CreateSession` alone in a single LogEntry, to avoid having 2 sessions with the same id.

common/condition_test.go
merlimat: Just saw this. I'll look into it.

server/public_rpc_server.go
andrasbeni: I'm working on this.

server/kv/db.go
andrasbeni: Yes, it's much cleaner to do it in a single callback. But then it needs to be one callback per operation type (Put, Delete, DeleteRange). 
I'll try to come up with an efficient way to find the session based on the key that is being deleted.

proto/storage.proto
```diff
@@ -10,3 +10,7 @@ message StorageEntry {
   fixed64 creation_timestamp = 3;
   fixed64 modification_timestamp = 4;
 }
+
+message SessionMetadata {
+  uint64 timeoutMS = 1;
```
merlimat: We are using uint32 in the create session request and uint64 here

server/kv/db.go
merlimat: Yes, I got confused with the status here, which is handled correctly

server/kv/db.go
```diff
@@ -253,6 +259,17 @@ func (d *db) applyPut(batch WriteBatch, notifications *notifications, putReq *pr
 	} else if err != nil {
 		return nil, errors.Wrap(err, "oxia db: failed to apply batch")
 	} else {
+		if updateOperationCallback != nil {
```
merlimat: Instead of always checking for `nil`, we could instead pass a `NoOpCallback` when we don't want to use it.

server/leader_controller.go
```diff
@@ -328,6 +339,15 @@ func (lc *leaderController) applyAllEntriesIntoDB() error {
 		return err
 	}
 
+	err = lc.sessionManager.Initialize()
+	if err != nil {
+		lc.log.Error().Err(err).
+			Int64("epoch", lc.epoch).
+			Uint32("shard", lc.shardId).
```
merlimat: Epoch & shardId are already attached to the contextual logger object

server/leader_controller.go
```diff
@@ -551,6 +575,7 @@ func (lc *leaderController) Close() error {
 	err = multierr.Combine(err,
 		lc.wal.Close(),
 		lc.db.Close(),
+		lc.sessionManager.Close(),
```
merlimat: Should we close it first since it depends on the DB?

server/public_rpc_server.go
```diff
@@ -125,6 +114,89 @@ func (s *PublicRpcServer) GetNotifications(req *proto.NotificationsRequest, stre
...
+	md, ok := metadata.FromIncomingContext(stream.Context())
+	if !ok {
+		return errors.New("shard id is not set in the request metadata")
+	}
+
+	shardId, err := ReadHeaderUint32(md, MetadataShardId)
+	if err != nil {
+		return err
+	}
+	sessionId, err := ReadHeaderUint64(md, "session-id")
```
merlimat: We should define a constant for `session-id` that is shared between client & server components

server/leader_controller.go
```diff
@@ -40,6 +41,10 @@ type LeaderController interface {
 
 	// Status The Status of the leader
 	Status() proto.ServingStatus
+
+	CreateSession(*proto.CreateSessionRequest) (*proto.CreateSessionResponse, error)
+	KeepAlive(shardId uint32, sessionId uint64, stream proto.OxiaClient_KeepAliveServer) error
```
merlimat: The `shardId` shouldn't be here since we already have the leader controller for a given shard

standalone/standalone_rpc_server.go
```diff
@@ -118,8 +124,79 @@ func (s *StandaloneRpcServer) Write(ctx context.Context, write *proto.WriteReque
...
 	lc, ok := s.controllers[*read.ShardId]
 	if !ok {
-		return nil, errors.New("shard not found")
+		return nil, ErrorShardNotFound
 	}
 
 	return lc.Read(read)
 }
+
+func (s *StandaloneRpcServer) CreateSession(ctx context.Context, req *proto.CreateSessionRequest) (*proto.CreateSessionResponse, error) {
```
merlimat: At some point we should just use the standard `PublicRpcServer` implementation for standalone too..  :)

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+)
+
+const (
+	KeyPrefix  = common.InternalKeyPrefix + "session/"
+	MaxTimeout = 10 * time.Second
+)
+
+var ErrorSessionNotFound = errors.New("session not found")
+
+var ErrorInvalidSessionTimeout = errors.New("invalid session timeout")
```
merlimat: (Pending merge of #170)

We should have these as constant gRPC codes like: https://github.com/streamnative/oxia/pull/170/files#diff-86ffcb900ecffb646ab3ba981cf58576fcae549b9515607d5d1127b81bd5db5fR16 

So that we can detect the specific problem in the client lib.

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+	sessions       map[SessionId]*session
+	sessionCounter SessionId
+	// channels to close heartbeat listeners
+	closeChannels []chan error
+	log           zerolog.Logger
+}
+
+func NewSessionManager(shardId uint32, controller LeaderController) SessionManager {
+	return &sessionManager{
+		Mutex:          sync.Mutex{},
```
merlimat: nit: no need to explicitly initialize the mutex

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+		sessions:       make(map[SessionId]*session),
+		sessionCounter: SessionId(0),
+		closeChannels:  make([]chan error, 0),
+		shardId:        shardId,
+		controller:     controller,
+		log: log.With().
+			Str("component", "session-manager").
```
merlimat, line 73: Also add `shard` to the logger

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+	"oxia/proto"
+	"oxia/server/kv"
+	"strconv"
+	"sync"
+	"time"
+)
+
+const (
+	KeyPrefix  = common.InternalKeyPrefix + "session/"
+	MaxTimeout = 10 * time.Second
```
merlimat: This should be more in the order of 5 minutes

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+func (sm *sessionManager) CreateSession(request *proto.CreateSessionRequest) (*proto.CreateSessionResponse, error) {
+	timeout := time.Duration(request.SessionTimeoutMs) * time.Millisecond
+	if timeout > MaxTimeout {
+		return nil, errors.Wrap(ErrorInvalidSessionTimeout, fmt.Sprintf("timeoutMs=%d", request.SessionTimeoutMs))
+	}
+	sm.Lock()
+	defer sm.Unlock()
+
+	sm.sessionCounter++
+	sessionId := sm.sessionCounter
```
merlimat: This can lead to repeated session ids after a new leader is elected.

I still think we need to tie the session id to the offset (and also change the type to `int64`)

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+	KeepAlive(shardId uint32, sessionId uint64, stream proto.OxiaClient_KeepAliveServer) error
+	CloseSession(*proto.CloseSessionRequest) (*proto.CloseSessionResponse, error)
+	Initialize() error
+}
+
+var _ SessionManager = (*sessionManager)(nil)
+
+type sessionManager struct {
+	sync.Mutex
+	controller     LeaderController
```
merlimat: I was a bit confused through the code at to what `controller` referred to. Can we rename to `leaderController`?

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+			Key:     SessionKey(sessionId),
+			Payload: marshalledMetadata,
+		}},
+	})
+	if err != nil || resp.Puts[0].Status != proto.Status_OK {
+		if err == nil {
+			err = errors.New("failed to register session")
+		}
+		return nil, err
+	}
```
merlimat: We should wrap the error here to provide more meaningful message: 

```suggestion
	if err != nil { 
            return nil, errors.Wrap(err, "failed to register session")
	}
	
	if resp.Puts[0].Status != proto.Status_OK {
            return nil, errors.Errorf("failed to register session. invalid status %#v", resp.Puts[0].Status)
	}
```

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+
+	result := map[SessionId]*proto.SessionMetadata{}
+
+	for i, metaEntry := range getResp.Gets {
+		key := gets[i].Key
+		if metaEntry.Status != proto.Status_OK {
+			sm.log.Warn().
+				Uint32("shard", sm.shardId).
```
merlimat: Shard should be included when the logger is created so we don't repeat it.

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+				Msgf("error reading session metadata at `%s`, status: %d", key, metaEntry.Status)
```
merlimat: We should prefer adding structured arguments instead of formatting

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+		if errors.Is(err, kv.ErrorKeyNotFound) {
+			return proto.Status_SESSION_DOES_NOT_EXIST, nil
+		}
+		return proto.Status_SESSION_DOES_NOT_EXIST, err
+	}
+	if err = closer.Close(); err != nil {
+		return proto.Status_SESSION_DOES_NOT_EXIST, err
+	}
+	// Create the session shadow entry
+	err = batch.Put(SessionKey(SessionId(*sessionId))+url.PathEscape(request.Key), []byte{})
```
merlimat: Couldn't we use the key directly here?

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+				StartInclusive: sessionKey,
+				EndExclusive:   sessionKey + "/",
+			},
+		},
+	})
+	return err
+
+}
+
+func (s *session) heartbeat(heartbeat *proto.SessionHeartbeat) {
```
merlimat: Since it's already a lot of code, we could move all the `session` related methods in a new file

server/session_manager.go
```diff
@@ -0,0 +1,513 @@
...
+	return s
+
+}
+
+func (s *session) waitForHeartbeats() {
+	s.log.Debug().Msg("Waiting for heartbeats")
+	timeout := s.timeout
+	for {
+		select {
+		case heartbeat := <-s.heartbeatCh:
```
merlimat: The heartbit here should reset the timeout, otherwise it will always expire.

standalone/standalone_rpc_server.go
andrasbeni: I couldn't agree more. Created a [backlog item](https://www.notion.so/streamnativeio/a8a710bff405448b911b45c2349ebdf3?v=1c376b03b96d4293ab5f84dcd735973f&p=2f38130d6fe2440a99edd85431358303&pm=s)

server/leader_controller.go
andrasbeni: `Close()` itself does not depend on the DB, but it's better to be cautious.

server/session_manager.go
andrasbeni: Yes, I removed this counter because I use the offset.

standalone/standalone_rpc_server.go
merlimat: Got ahead of it :) https://github.com/streamnative/oxia/pull/185

proto/storage.proto
```diff
@@ -9,4 +9,9 @@ message StorageEntry {
   int64 version = 2;
   fixed64 creation_timestamp = 3;
   fixed64 modification_timestamp = 4;
+  optional uint64 sessionId = 5;
+}
+
+message SessionMetadata {
+  uint32 timeoutMS = 1;
```
merlimat: nit: we're using `_` for all proto naming
```suggestion
  uint32 timeout_ms = 1;
```

server/errors.go
```diff
@@ -10,6 +10,7 @@ const (
 	CodeInvalidEpoch    codes.Code = 100
 	CodeInvalidStatus   codes.Code = 101
 	CodeNotServingShard codes.Code = 102
+	CodeInvalidSession  codes.Code = 103
```
merlimat: This file has already moved to `common/error_codes.go`

server/leader_controller.go
```diff
@@ -21,7 +21,10 @@ type LeaderController interface {
 	io.Closer
 
 	Write(write *proto.WriteRequest) (*proto.WriteResponse, error)
+	write(request func(int64) *proto.WriteRequest) (int64, *proto.WriteResponse, error)
```
merlimat: private methods shouldn't be added to the interface. We can pass around a `*leaderController` if we need to use them from outside

server/session.go
```diff
@@ -0,0 +1,155 @@
...
+		select {
+
+		case heartbeat := <-heartbeatChannel:
+			if heartbeat == nil {
+				// The channel is closed, so the session must be closing
+				return
+			}
+		case <-time.After(timeout):
+			s.log.Info().
+				Msg("Session timed out")
```
merlimat: Session expired

proto/client.proto
```diff
@@ -176,6 +191,9 @@ message PutRequest {
   // An optional expected version. The put will fail if the server's current version
   // does not match
   optional int64 expected_version = 3;
+  // Optional. Associate the new value with the session (i.e. ephemeral value).
+  // When the session times out or is closed, the key-payload pair will be removed
+  optional uint64 session_id = 4;
```
merlimat: I'd say to use `int64` for session ids as well, to be consistent with offset type

server/session_manager.go
```diff
@@ -0,0 +1,307 @@
...
+			Key:            key,
+			IncludePayload: true,
+		})
+	}
+	getResp, err := sm.controller.readWithoutLocking(&proto.ReadRequest{
+		ShardId: &sm.shardId,
+		Gets:    gets,
+		Lists:   nil,
+	})
+	if err != nil {
```
merlimat, line 204: Separate issue, though I feel we should add a better way to do this in the DB interface

server/session_manager.go
```diff
@@ -0,0 +1,307 @@
...
+				Msgf("error reading session metadata")
+			continue
+		}
+		payload := metaEntry.Payload
+		metadata := proto.SessionMetadata{}
+		err = pb.Unmarshal(payload, &metadata)
+		if err != nil {
+			sm.log.Warn().
+				Err(err).
+				Str("key", key).
```
merlimat, line 223: Should we log the sessionId as well?

server/session_manager.go
```diff
@@ -0,0 +1,307 @@
...
+				Msgf("error unmarshalling session metadata")
+			continue
+		}
+		sessionId, err := KeyToId(key)
+		if err != nil {
+			sm.log.Warn().
+				Err(err).
+				Str("key", key).
```
merlimat, line 234: Same here, logging the session id

server/session_manager.go
```diff
@@ -0,0 +1,307 @@
...
+
+type updateCallback struct{}
+
+var SessionUpdateOperationCallback kv.UpdateOperationCallback = &updateCallback{}
+
+func (_ *updateCallback) OnPut(batch kv.WriteBatch, request *proto.PutRequest, existingEntry *proto.StorageEntry) (proto.Status, error) {
+	if existingEntry != nil && existingEntry.SessionId != nil {
+		// We are overwriting an ephemeral value, let's delete its shadow
+		status, err := deleteShadow(batch, request.Key, existingEntry)
+		if err != nil {
```
merlimat: One less line: 

```suggestion
		if status, err := deleteShadow(batch, request.Key, existingEntry); err != nil {
```

proto/client.proto
```diff
@@ -278,6 +296,8 @@ message Stat {
   fixed64 created_timestamp = 2;
   // The modified timestamp of the current version of the key
   fixed64 modified_timestamp = 3;
+  // Identifier of the session if this Stat object belongs to an ephemeral value
+  uint64 session_id = 4;
```
merlimat: Either we make this optional or we use -1 to signal "no session"

server/session_manager.go
```diff
@@ -0,0 +1,307 @@
...
+			if errors.Is(err, kv.ErrorKeyNotFound) {
+				return proto.Status_SESSION_DOES_NOT_EXIST, nil
+			}
+			return proto.Status_SESSION_DOES_NOT_EXIST, err
+		}
+		if err = closer.Close(); err != nil {
+			return proto.Status_SESSION_DOES_NOT_EXIST, err
+		}
+		// Create the session shadow entry
+		err = batch.Put(SessionKey(SessionId(*sessionId))+url.PathEscape(request.Key), []byte{})
```
merlimat, line 284: I think my previous comment here got lost in the split. Why do we need to escape the key? Can't we just use it as it is?

server/session_manager.go
andrasbeni: We can't log it here because we don't have it because parsing in `KeyToId` failed. Our best information is the DB key which hopefully will help us figure out which session's ID is corrupted or what other info is put in the wrong place.

server/session_manager.go
andrasbeni: I use the range: `StartInclusive: SessionKey(...), EndExclusive: SessionKey(...)+"/"` to read all shadow nodes. This only works as long as the child's name does not have a `/` in it.

In other words:
`__oxia/session/$session_id/  < __oxia/session/$session_id/a%2Fb < __oxia/session/$session_id//` but
`__oxia/session/$session_id/  < __oxia/session/$session_id//     < __oxia/session/$session_id/a/b`

Is there a better workaround for this?

server/session_manager.go
andrasbeni: Resolving this conversation in favor of the one started on `session.go`

server/session_manager.go
andrasbeni: I'm not sure what you mean, can you please elaborate?

proto/client.proto
merlimat:
```suggestion
  optional int64 session_id = 4;
```

proto/storage.proto
```diff
@@ -9,4 +9,9 @@ message StorageEntry {
   int64 version = 2;
   fixed64 creation_timestamp = 3;
   fixed64 modification_timestamp = 4;
+  optional uint64 sessionId = 5;
```
merlimat:
```suggestion
  optional int64 sessionId = 5;
```

server/session.go
```diff
@@ -0,0 +1,154 @@
...
+
+func startSession(sessionId SessionId, sessionMetadata *proto.SessionMetadata, sm *sessionManager) *session {
+	s := &session{
+		id:          sessionId,
+		timeout:     time.Duration(sessionMetadata.TimeoutMs) * time.Millisecond,
+		sm:          sm,
+		heartbeatCh: make(chan *proto.SessionHeartbeat, 1),
+		closeCh:     make(chan error),
+		log: sm.log.With().
+			Str("session-id", hexId(sessionId)).Logger(),
```
merlimat: Nit: Using base 10 when printing the session ids in logs (consistently with offsets/shard-id...)

```suggestion
			Int64("session-id", sessionId).Logger(),
```

server/session_manager.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func NewSessionManager(shardId uint32, controller *leaderController) SessionManager {
+	return &sessionManager{
+		sessions:         make(map[SessionId]*session),
+		shardId:          shardId,
+		leaderController: controller,
+		log: log.With().
+			Str("component", "session-manager").
+			Uint32("shard", shardId).
```
merlimat, line 74: Earlier I forgot: we could also include the epoch here

server/session_manager.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	result := map[SessionId]*proto.SessionMetadata{}
+
+	for i, metaEntry := range getResp.Gets {
+		key := gets[i].Key
+		if metaEntry.Status != proto.Status_OK {
+			sm.log.Warn().
+				Str("key", key).
+				Str("status", fmt.Sprintf("%#v", metaEntry.Status)).
```
merlimat:
```suggestion
				Stringer("status", metaEntry.Status).
```

server/session_manager.go
```diff
@@ -0,0 +1,305 @@
...
+				Msgf("error parsing session key")
+			continue
+		}
+		payload := metaEntry.Payload
+		metadata := proto.SessionMetadata{}
+		err = pb.Unmarshal(payload, &metadata)
+		if err != nil {
+			sm.log.Warn().
+				Err(err).
+				Int32("session-id", int32(sessionId)).
```
merlimat:
```suggestion
				Int64("session-id", sessionId).
```

server/session_manager.go
merlimat: I guess we'd have the problem only with the first character. 

Could it work with `EndExclusive: SessionKey(...)+"\xFF"` ?

server/session.go
```diff
@@ -0,0 +1,177 @@
...
+		timeout:     time.Duration(sessionMetadata.TimeoutMs) * time.Millisecond,
+		sm:          sm,
+		heartbeatCh: make(chan *proto.SessionHeartbeat, 1),
+		closeCh:     make(chan error),
+		log: sm.log.With().
+			Str("component", "session").
+			Str("session-id", hexId(sessionId)).Logger(),
+	}
+	go s.waitForHeartbeats()
+	s.log.Info().Msg("Session started")
```
merlimat: I'd use debug here. If there are many session it will print a lot of logs each time a new leader is elected.

server/session_manager.go
merlimat: Nothing, just the feeling that we could have better interface in DB to do a scan of (key, values) in one pass, instead of having to fetch the list and then read all of them. way out of scope for this PR :)

andrasbeni: Yes, I was surprised I could not do it in one go.

PR #123, opened 2022-12-13
Remove epoch from RPC responses

merlimat:
Stacked on top of #114, #116 and #122

Since Grpc is already correlating the request & response pairs, there is no use to have an "epoch" field in the response.

Added `EmptyResponse` message to use when no response fields are required.

### DIFF
--- a/server/leader_controller_test.go
+++ b/server/leader_controller_test.go
@@ -417,7 +417,7 @@ func TestLeaderController_BecomeLeaderEpoch(t *testing.T) {
 	assert.Equal(t, CodeInvalidEpoch, status.Code(err))
 
 	// Same epoch will succeed
-	resp, err = lc.BecomeLeader(&proto.BecomeLeaderRequest{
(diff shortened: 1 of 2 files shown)


proto/coordination.proto
```diff
@@ -8,19 +8,20 @@ import "proto/client.proto";
 
 // controller -> node
 service OxiaControl {
-  rpc ShardAssignment(stream io.streamnative.oxia.proto.ShardAssignmentsResponse) returns (CoordinationShardAssignmentsResponse) {}
+  rpc ShardAssignment(stream io.streamnative.oxia.proto.ShardAssignmentsResponse) returns (EmptyResponse);
```
andrasbeni: Using the same message type for different responses couples together those RPCs and they cannot evolve separately.

merlimat: Good point. I don't see anything that we could add, though I agree it's better not to tie our hands unnecessarily here

PR #125, opened 2022-12-14
Added ability to create Db snapshots

merlimat:
Changed the interface for exposing DB snapshots. 

Instead of having an iterator over the (key,value) pair, we expose the snapshot as a sequence of files. 

This has the advantage that we can transfer much faster a set of compressed SST files and re-create the database in a different node, while the iterator approach will have to decompress all the values and stream them to the other server.

### DIFF
--- a/server/leader_controller_test.go
+++ b/server/leader_controller_test.go
@@ -446,10 +446,11 @@ func TestLeaderController_BecomeLeaderEpoch(t *testing.T) {
 func TestLeaderController_EntryVisibilityAfterBecomingLeader(t *testing.T) {
 	var shard uint32 = 1
 
-	kvFactory := kv.NewPebbleKVFactory(&kv.KVFactoryOptions{
+	kvFactory, err := kv.NewPebbleKVFactory(&kv.KVFactoryOptions{
 		DataDir:   t.TempDir(),
 		CacheSize: 10 * 1024,
 	})
+	assert.NoError(t, err)
 	walFactory := wal.NewWalFactory(&wal.WalFactoryOptions{
 		LogDir: t.TempDir(),
 	})


server/kv/kv_pebble.go
```diff
@@ -32,15 +34,37 @@ func NewPebbleKVFactory(options *KVFactoryOptions) KVFactory {
 		dataDir = DefaultKVFactoryOptions.DataDir
 	}
 
+	// Cleanup leftover snapshots from previous runs
```
andrasbeni: This comment belongs to line 49, I believe.

server/kv/kv_pebble.go
```diff
@@ -282,3 +319,120 @@ func CompareWithSlash(a, b []byte) int {
...
+
+func newPebbleSnapshot(p *Pebble) (Snapshot, error) {
+	ps := &pebbleSnapshot{
+		path: filepath.Join(p.dataDir, "snapshots",
+			fmt.Sprintf("shard-%d", p.shardId),
+			fmt.Sprintf("snapshot-%d", p.snapshotCounter.Add(1))),
+	}
+
+	// Flush the DB to ensure the snapshot content is as close as possible to
+	// the head
```
andrasbeni: We can't get closer to head than `commitIndex`, and even if we could, we would not want to, so I think it's better to name that as our goal:
```suggestion
	// the last committed entry 
```

server/kv/kv_pebble.go
```diff
@@ -32,15 +34,37 @@ func NewPebbleKVFactory(options *KVFactoryOptions) KVFactory {
...
-	return &PebbleFactory{
+	pf := &PebbleFactory{
 		dataDir: dataDir,
 		options: options,
 
 		// Share a single cache instance across the databases for all the shards
 		cache: cache,
 	}
+
+	if err := pf.cleanupSnapshots(); err != nil {
```
andrasbeni, line 48: Do we need to stop the whole worker because of this error? Sure, it's important, but I think an error log (and later incrementing an error counter that's exposed through metrics) will suffice. And then, we could just increment the `snapshotCounter` until we find a directory that does not yet exist.

server/kv/db.go
```diff
@@ -28,6 +28,8 @@ type DB interface {
 
 	UpdateEpoch(newEpoch int64) error
 	ReadEpoch() (epoch int64, err error)
+
+	KV() KV
```
andrasbeni: Do you have any reason to expose the `KV` here other than snapshots? If not, I'd only expose a method to create snapshots.

server/kv/kv.go
```diff
@@ -36,6 +36,29 @@ type KeyValueIterator interface {
 	Value() ([]byte, error)
 }
 
+type SnapshotFile interface {
+	Path() string
+	Content() ([]byte, error)
+}
```
andrasbeni: I don't think we need to be explicit about using files. 
```suggestion
type SnapshotChunk interface {
	Name() string
	Content() ([]byte, error)
}
```

server/kv/db.go
merlimat: Yes, we don't even really need it. I'm removing this

server/kv/kv_pebble.go
merlimat: If we're not able to delete that directory, it's probably a symptom of a more significant issue. This is in the same data directory where all the dbs are stored. It could be a permission or disk issue that will most likely affect everything else. 

> Do we need to stop the whole worker because of this error?  

It will not stop a working server. This is done at startup and it will prevent starting. I think it's appropriate so that the issue can be properly investigated.

andrasbeni: 👍

PR #131, opened 2022-12-15
Fix cluster issues

nahguam:


server/container/container.go
```diff
@@ -28,7 +28,7 @@ func Start(name string, port int, registerFunc func(grpc.ServiceRegistrar)) (*Co
 	registerFunc(c.server)
 	grpc_prometheus.Register(c.server)
 
-	listener, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
```
merlimat, line 31: Eventually, we should be taking a bind address instead of just the port as a config argument

PR #132, opened 2022-12-15
Follower controller handling of snapshot

merlimat:
Added the `FollowerController` portion of handling the shipment of a snapshot from a leader to a follower.

The snapshot is not changing the nature of the replication protocol, instead it allows for a fast forward option that the leader can choose in a few cases : 

 * Bootstrapping a follower that has no data (perhaps the follower disk had been wiped)
 * Fast-forwarding a follower that comes back and it's very far behind the leader

### DIFF
--- a/server/internal_rpc_server.go
+++ b/server/internal_rpc_server.go
@@ -228,7 +228,7 @@ func (s *internalRpcServer) AddEntries(srv proto.OxiaLogReplication_AddEntriesSe
 }
 
 func (s *internalRpcServer) SendSnapshot(srv proto.OxiaLogReplication_SendSnapshotServer) error {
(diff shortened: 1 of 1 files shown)


server/internal_rpc_server.go
```diff
@@ -227,6 +227,42 @@ func (s *internalRpcServer) AddEntries(srv proto.OxiaLogReplication_AddEntriesSe
 	}
 }
 
+func (s *internalRpcServer) SendSnapshot(srv proto.OxiaLogReplication_SendSnapshotServer) error {
+	// Add entries receives an incoming stream of request, the shard_id needs to be encoded
```
andrasbeni:
```suggestion
	// Send snapshot receives an incoming stream of requests, the shard_id needs to be encoded
```

PR #133, opened 2022-12-15
Fixed assignment dispatcher disconnection

merlimat:
Fixing a couple of issues on the server-side component that dispatches the assignment updates: 

 * The Grpc handler methods must wait until the client stream is open. Right now, we're sending the first response back to the client and then returning, which closes the stream
 * The client stream is staying on the `clients` map until we try to dispatch a new assignment change. This means if we have many clients connecting/disconnecting, the map will grow unbound
 * When dispatching updates to clients, we shouldn't be blocked if one client is slow/unresponsive

### DIFF
--- a/server/assignment_dispatcher.go
+++ b/server/assignment_dispatcher.go
@@ -55,6 +55,7 @@ func (s *shardAssignmentDispatcher) RegisterForUpdates(clientStream Client) erro
 	s.nextClientId++
 
 	s.clients[clientId] = clientCh
+	closeCh := s.closeCh
 	s.Unlock()
 
 	// Send initial assignments
@@ -93,7 +94,7 @@ func (s *shardAssignmentDispatcher) RegisterForUpdates(clientStream Client) erro
 			s.Unlock()
 			return nil
 
-		case <-s.closeCh:
+		case <-closeCh:
 			// the server is closing
 			return ErrorAlreadyClosed
 		}


server/assignment_dispatcher.go
```diff
@@ -23,34 +25,75 @@ type ShardAssignmentsDispatcher interface {
...
+
+	s.clients[clientId] = clientCh
+	s.Unlock()
+
+	// Send initial assignments
 	err := clientStream.Send(s.assignments)
 	if err != nil {
+		s.Unlock()
 		return err
 	}
```
andrasbeni: `s.assignments` needs to be guarded, so moving the Unlock call to after its usage and removing the redundant Unlock would make sense, IMO:
```suggestion

	// Send initial assignments
	err := clientStream.Send(s.assignments)
	s.Unlock()
	if err != nil {
		return err
	}
```

server/assignment_dispatcher.go
```diff
@@ -112,13 +155,15 @@ func (s *shardAssignmentDispatcher) updateShardAssignment(assignments *proto.Sha
...
-				Str("client", peer.Addr.String()).
-				Msg("Failed to send shard assignment update to client")
+	// Update all the clients, without getting stuck if any client is not responsive
+	for id, clientCh := range s.clients {
+		select {
+		case clientCh <- assignments:
+			// Good, we were able to pass the update to the client
+
+		default:
+			// The client is not responsive, cut him off
```
andrasbeni: nit
```suggestion
			// The client is not responsive, cut it off
```

server/assignment_dispatcher.go
merlimat: 👍 Fixed the locking

server/assignment_dispatcher.go
```diff
@@ -23,35 +25,80 @@ type ShardAssignmentsDispatcher interface {
...
+			if err != nil {
+				if status.Code(err) != codes.Canceled {
+					peer, _ := peer.FromContext(clientStream.Context())
+					s.log.Warn().Err(err).
+						Str("client", peer.Addr.String()).
+						Msg("Failed to send shard assignment update to client")
+				}
+				s.Lock()
+				delete(s.clients, clientId)
+				s.Unlock()
```
merlimat, line 87: Uhm.. I was missing a return here :/

PR #135, opened 2022-12-16
Bind address

nahguam:
Stacked on #134 

True diff: https://github.com/streamnative/oxia/compare/sync-cluster-chart-go...bind-address

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -1,13 +1,13 @@
 package coordinator
 
 import (
+	"fmt"
 	"github.com/rs/zerolog/log"
 	"go.uber.org/multierr"
 	"oxia/common"
 	"oxia/common/metrics"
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -1,15 +1,17 @@
...
 	"go.uber.org/multierr"
 	"oxia/common"
+	"oxia/common/metrics"
 	"oxia/server/kv"
-	"oxia/server/metrics"
 	"oxia/server/wal"
 )
 
 type Config struct {
+	BindHost            string
```
merlimat, line 14: To make it more flexible, we could take `PublicServiceAddress=localhost:6648` or ``PublicServiceAddress=0.0.0.0:6648` so that each service could bind on different addresses. Eg: public service on 0.0.0.0 but metrics on localhost only.

PR #136, opened 2022-12-16
Configmap metadataprovider

nahguam:
Stacked on #135 

True diff: https://github.com/streamnative/oxia/compare/bind-address...configmap-metadataprovider

### DIFF
--- a/coordinator/impl/metadata_configmap.go
+++ b/coordinator/impl/metadata_configmap.go
@@ -1,0 +1,98 @@
+package impl
+
+import (
+	"github.com/rs/zerolog/log"
+	"gopkg.in/yaml.v2"
+	coreV1 "k8s.io/api/core/v1"
+	k8sError "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"oxia/coordinator/model"
+	"oxia/kubernetes"
+	"sync"
+)
+
+type metadataProviderConfigMap struct {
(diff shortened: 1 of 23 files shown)


coordinator/impl/metadata_configmap.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if version != expectedVersion {
+		err = ErrUnexpectedVersion
+		return
+	}
+
+	version++
+
+	err = kubernetes.ConfigMaps(m.kubernetes).Upsert(m.namespace, configMap(m.name, status, version))
```
merlimat: If the version check is not supported by the ConfigMap API, we would have a race condition here if there are 2 coordinator processes running at the same time, and we would lose some of the updates

nahguam: Good point. I believe there might be a revision property that can be checked.

nahguam: There isn't on cm resources. But even if there were, I don't see a way to atomically update the resource. The same is true on CRD status. Actually not sure whether the `resourceVersion` applies there. Will need to read up a bit more

nahguam: @merlimat Yes we can use the `resourceVersion`, as long as it's treated as opaque. It currently maps to the etcd equivalent but they might have plans to change that in the future. It's actually represented as a `string` in the model, but parsed to a `uint` internally. 

I'll change the data type to a `string` to match the k8s go-client API:

`type Version string`

merlimat: 👍

PR #142, opened 2022-12-17
Fixed follower replay from commit index

merlimat:
Fixed 2 issues in follower controller: 
 1. It should recover the commit index after restart
 2. When it receives an update with a commit-index from the leader that was higher than its own commit-index, it was blindly updating it, even though we hadn't put these entries yet on the db

Also added more debug logs to make sense of the sequence of events.

### DIFF
--- a/server/assignment_dispatcher.go
+++ b/server/assignment_dispatcher.go
@@ -33,8 +33,11 @@ type shardAssignmentDispatcher struct {
 	assignments  *proto.ShardAssignmentsResponse
 	clients      map[int64]chan *proto.ShardAssignmentsResponse
 	nextClientId int64
-	closeCh      chan error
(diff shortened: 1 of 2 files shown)


server/follower_controller.go
```diff
@@ -172,6 +176,12 @@ func (fc *followerController) Epoch() int64 {
 	return fc.epoch
 }
 
+func (fc *followerController) CommitIndex() int64 {
+	fc.Lock()
+	defer fc.Unlock()
+	return fc.commitIndex
```
nahguam, line 182: Does this need a lock?

merlimat: As it is yes, because it's changed by different go-routine, though we should probably change all these variables with `atomic.Int64` so we don't need the locks

PR #145, opened 2022-12-19
Shard upperbound inclusive

nahguam:


server/assignment_dispatcher_test.go
```diff
@@ -138,8 +138,28 @@ func newShardAssignment(id uint32, leader string, min uint32, max uint32) *proto
...
 		},
 	}
 }
+
+func TestShardGenerator(t *testing.T) {
+	assignments := generateShards("localhost", 4)
+	assertNext(t, assignments[0], 0, 0, 1073741823)
+	assertNext(t, assignments[1], 1, 1073741824, 2147483647)
+	assertNext(t, assignments[2], 2, 2147483648, 3221225471)
+	assertNext(t, assignments[3], 3, 3221225472, 4294967295)
```
merlimat: For visual assurance we could use hex base here :), or use `math.MaxUint64` instead of `4294967295`

PR #146, opened 2022-12-20
Support configuration of metadata provider

nahguam:


cmd/coordinator/cmd.go
```diff
@@ -17,18 +18,36 @@ var (
...
+		PreRunE: validate,
 		Run:     exec,
 	}
 )
 
 func init() {
 	flag.InternalPort(Cmd, &conf.InternalServicePort)
 	flag.MetricsPort(Cmd, &conf.MetricsPort)
+	Cmd.Flags().Var(&conf.MetadataProviderImpl, "metadata", "Metadata provider implementation: memory or configmap")
+	Cmd.Flags().StringVar(&conf.MetadataNamespace, "namespace", conf.MetadataNamespace, "Kubernetes namespace for metadata configmap")
```
merlimat: Maybe we should add `--k8s-` prefix for these settings. eg.  `--k8s-namespace` and `--k8s-configmap-name`

nahguam: yes, good call. I'll update

PR #147, opened 2022-12-20
Add retry with backoff to client side shard assignment

nahguam:


oxia/internal/shard_manager.go
```diff
@@ -114,15 +116,29 @@ func (s *shardManagerImpl) isClosed() bool {
 }
 
 func (s *shardManagerImpl) receiveWithRecovery(ctx context.Context, readyC chan bool) {
-	for {
-		err := s.receive(ctx, readyC)
-		if err != nil && s.isClosed() {
-			return
-		}
+	backOff := common.NewBackOff(context.Background())
```
merlimat:
```suggestion
	backOff := common.NewBackOff(ctx)
```

merlimat: Shouldn't this be tied to the same context?

nahguam: I was initially going to do that, but it's also being used for the underlying rpc call. Should they be shared like that?

nahguam: How about we use the provided one here as you suggest and use `context.Background()` for the rpc call?

merlimat: Since the retry for receiving the shard assignments is tied to the lifecycle of the client instance, we should probably use a context  from the client instance. When the client is closed, we cancel the context and we can stop the retry

nahguam: Yes, the `ctx` coming into the function is the one created in `Start()` and it's cancelled on close so it is indirectly already bound to the lifecycle of the client. I'll go with your original suggestion and use it. And I'll also leave the underlying rpc call to use the same one.

PR #150, opened 2022-12-21
Make the Grpc provider pluggable for Oxia server

merlimat:
Allowing to pass client & server provider for abstracting Grpc transport in server.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -49,6 +49,7 @@ func NewWithGrpcProvider(config Config, provider container.GrpcProvider, replica
 	}
 
 	s := &Server{
+		replicationRpcProvider: replicationRpcProvider,
 		walFactory: wal.NewWalFactory(&wal.WalFactoryOptions{
 			LogDir: config.WalDir,
 		}),


common/container/container.go
```diff
@@ -10,15 +10,34 @@ import (
 	"oxia/common"
 )
 
-type Container struct {
+type GrpcServer interface {
```
andrasbeni, line 13: At this point, I think it's worth renaming the file to e.g. `grpc_server.go`

standalone/standalone_rpc_server.go
```diff
@@ -37,7 +36,7 @@ func NewStandaloneRpcServer(bindAddress string, advertisedPublicAddress string,
 		numShards:               numShards,
 		walFactory:              walFactory,
 		kvFactory:               kvFactory,
-		clientPool:              common.NewClientPool(),
+		replicationRpcProvider:  server.NewReplicationRpcProvider(),
```
andrasbeni, line 39: How about: 
```suggestion
		replicationRpcProvider:  nil,
```
I haven't tested it, but if this does not work, then I think we're doing something wrong.

merlimat: I'm not sure I'm following this change. If we don't initialize it, we'll be passing a nil to the `LeaderController`. It won't work, but I don't understand why would that be wrong.

andrasbeni: This is the standalone server, so it must not try to replicate anything.

merlimat: Oh, it's true. Though it will crash when calling Close()

PR #152, opened 2022-12-21
Added file based metadata impl

merlimat:
Added option to use file-based metadata for the coordinator (in addition to config-map and in-memory). This is going to be mostly used for local testing.


cmd/coordinator/cmd.go
```diff
@@ -26,18 +26,19 @@ var (
 func init() {
 	flag.InternalPort(Cmd, &conf.InternalServicePort)
 	flag.MetricsPort(Cmd, &conf.MetricsPort)
-	Cmd.Flags().Var(&conf.MetadataProviderImpl, "metadata", "Metadata provider implementation: memory or configmap")
-	Cmd.Flags().StringVar(&conf.MetadataNamespace, "k8s-namespace", conf.MetadataNamespace, "Kubernetes namespace for metadata configmap")
-	Cmd.Flags().StringVar(&conf.MetadataName, "k8s-configmap-name", conf.MetadataName, "ConfigMap name for metadata configmap")
+	Cmd.Flags().Var(&conf.MetadataProviderImpl, "metadata", "Metadata provider implementation: file, configmap or memory")
+	Cmd.Flags().StringVar(&conf.K8SMetadataNamespace, "k8s-namespace", conf.K8SMetadataNamespace, "Kubernetes namespace for metadata configmap")
+	Cmd.Flags().StringVar(&conf.K8SMetadataConfigMapName, "k8s-configmap-name", conf.K8SMetadataConfigMapName, "ConfigMap name for metadata configmap")
+	Cmd.Flags().StringVar(&conf.FileMetadataPath, "file-clusters-status-path", "data/cluster-status.json", "The path where the store the cluster status when using 'file' provider")
```
andrasbeni:
```suggestion
	Cmd.Flags().StringVar(&conf.FileMetadataPath, "file-clusters-status-path", "data/cluster-status.json", "The path where the cluster status is stored when using 'file' provider")
```

cmd/coordinator/cmd.go
```diff
@@ -26,18 +26,19 @@ var (
...
 	if conf.MetadataProviderImpl == coordinator.Configmap {
-		if conf.MetadataNamespace == "" {
+		if conf.K8SMetadataNamespace == "" {
 			return errors.New("k8s-namespace must be set with metadata=configmap")
 		}
-		if conf.MetadataName == "" {
+		if conf.K8SMetadataConfigMapName == "" {
 			return errors.New("k8s-configmap-name must be set with metadata=configmap")
 		}
 	}
```
andrasbeni, line 44: I'm not sure if this would help anyone, but maybe it's still better to be strict.
```suggestion
	}
	if conf.MetadataProviderImpl == coordinator.File || conf.MetadataProviderImpl == coordinator.Memory {
		if conf.K8SMetadataNamespace != "" {
			return errors.New("k8s-namespace must only be set with metadata=configmap")
		}
		if conf.K8SMetadataConfigMapName != "" {
			return errors.New("k8s-configmap-name must only be set with metadata=configmap")
		}
	}
```

coordinator/impl/metadata_file.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+		} else {
+			return MetadataNotExists, err
+		}
+	}
+
+	if err := m.fileLock.Lock(); err != nil {
+		return "", errors.Wrap(err, "failed to acquire file lock")
+	}
+	defer func() { _ = m.fileLock.Unlock() }()
```
andrasbeni: Should we not log this error?

cmd/coordinator/cmd_test.go
```diff
@@ -43,7 +43,7 @@ func TestCmd(t *testing.T) {
 		{[]string{}, coordinator.Config{
 			InternalServicePort:  6649,
 			MetricsPort:          8080,
-			MetadataProviderImpl: coordinator.Memory,
+			MetadataProviderImpl: coordinator.File,
```
andrasbeni, line 46: Is it intentional that you eliminate testing the Memory provider from here?

merlimat: I've changed the default from `Memory` -> `File`. The line above is to correct the test expecation.

cmd/coordinator/cmd.go
merlimat: Sure, though I'd say it's for a different PR

coordinator/impl/metadata_file.go
merlimat: Added log

PR #154, opened 2022-12-21
When coordinator restarts, we should leave the current leader

merlimat:


coordinator/impl/shard_controller_test.go
```diff
@@ -84,6 +84,37 @@ func TestShardController(t *testing.T) {
...
+
+	select {
+	case <-rpc.getNode(s1).fenceRequests:
+		assert.Fail(t, "shouldn't have received any fence requests")
+	case <-rpc.getNode(s2).fenceRequests:
+		assert.Fail(t, "shouldn't have received any fence requests")
+	case <-rpc.getNode(s3).fenceRequests:
+		assert.Fail(t, "shouldn't have received any fence requests")
+
+	case <-time.After(1 * time.Second):
```
andrasbeni, line 111: I'd love to suggest a solution that does not involve idling a second, but unfortunately, I can't come up with anything better.

PR #164, opened 2022-12-28
Notification support in DB

merlimat:
First part, added support in DB to handle the log of notifications.

As described in https://www.notion.so/streamnativeio/Notifications-4f1a9c5e09da4346a05cf27f6cf0f3b0

### DIFF
--- a/server/kv/notifications_tracker.go
+++ b/server/kv/notifications_tracker.go
@@ -1,6 +1,7 @@
 package kv
 
 import (
+	"context"
 	"fmt"
 	"github.com/pkg/errors"
 	pb "google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


server/kv/notifications_tracker.go
```diff
@@ -0,0 +1,116 @@
...
+		res = append(res, nb)
+	}
+
+	return res, nil
+}
+
+func (nt *notificationsTracker) Close() error {
+	nt.closed.Store(true)
+	return nil
+}
```
andrasbeni:
```suggestion
func (nt *notificationsTracker) Close() error {
	nt.closed.Store(true)
	nt.cond.Broadcast()
	return nil
}
```

PR #167, opened 2022-12-29
Added support for testing Oxia with Maelstrom

merlimat:
Added a new binary `oxia-maelstrom` that can does all the message exchanging through JSON frames on stdin/stdout. It can be used for testing Oxia through Maelstrom (https://github.com/jepsen-io/maelstrom).

The `oxia-maelstrom` binary will use special RPC providers that replace the gRPC calls and tie them to the JSON frames.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,7 @@ build: proto crd
 	go build -v -o bin/oxia ./cmd
 
 .PHONY: maelstrom
-maelstrom:
-	#protoc --grpchan_out=proto proto/coordination.proto
-#--go_out=plugins=grpc:maelstrom --grpchan_out=maelstrom proto/coordination.proto
-#--go_opt=paths=source_relative
+maelstrom: proto
 	go build -v -o bin/oxia-maelstrom ./maelstrom
 
 test: build
--- a/maelstrom/replication_rpc_provider.go
+++ b/maelstrom/replication_rpc_provider.go
@@ -64,6 +64,10 @@ func (r *maelstromReplicationRpcProvider) Truncate(follower string, req *proto.T
 	}
 }
 
+func (r *maelstromReplicationRpcProvider) SendSnapshot(ctx context.Context, follower string, shard uint32) (proto.OxiaLogReplication_SendSnapshotClient, error) {
+	panic("not implemented")
+}
+
 // //////// AddEntriesClient
 type maelstromAddEntriesClient struct {
 	BaseStream


maelstrom/main.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return LogLevelError(logLevelStr)
+	}
+	common.LogLevel = logLevel
+	common.ConfigureLogger()
+	return nil
+}
+
+var thisNode string
+var allNodes []string
```
andrasbeni, line 50: Could these be return values of `handleInit()` instead?

maelstrom/main.go
```diff
@@ -0,0 +1,147 @@
...
+		log.Fatal().Err(err).Msg("failed to create data dir")
+	}
+
+	if thisNode == "n1" {
+		// First node is going to be the "coordinator"
+		_, err := impl.NewCoordinator(
+			impl.NewMetadataProviderFile(filepath.Join(dataDir, "cluster-status.json")),
+			model.ClusterConfig{
+				ReplicationFactor: 3,
+				ShardCount:        1,
```
andrasbeni, line 126: It would be nice to be able to test with more than one shard. Is there any particular reason you decided to create only a single shard?

maelstrom/coordinator_rpc_client.go
```diff
@@ -0,0 +1,216 @@
...
+	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/metadata"
+	"os"
+	"oxia/coordinator/impl"
+	"oxia/coordinator/model"
+	"oxia/proto"
+	"sync"
+)
+
+type maelstromCoordinatorRpcProvider struct {
```
andrasbeni, line 17: The file name needs to be more accurate.
I think both `coordinator_rpc.go` and `coordinator_rpc_provider.go` could work.

maelstrom/messages.go
```diff
@@ -0,0 +1,276 @@
...
+			Str("line", line).
+			Msg("failed to unmarshal")
+	}
+
+	msg = frame
+	msgType = frame.Body.Type
+	sm, ok := jsonMsgMapping[frame.Body.Type]
+	if ok {
+		// Deserialize json again with specific type
+		// Deserialize again with the proper struct
```
andrasbeni, line 233: It looks like one of these comments is redundant.

maelstrom/messages.go
```diff
@@ -0,0 +1,276 @@
...
+	MsgTypeWrite   = "write"
+	MsgTypeWriteOk = "write_ok"
+	MsgTypeRead    = "read"
+	MsgTypeReadOk  = "read_ok"
+	MsgTypeCas     = "cas"
+	MsgTypeCasOk   = "cas_ok"
+
+	/* Oxia specific messages */
+
+	MsgTypeFenceRequest         = "fence-req"
```
andrasbeni: It would help maintainability if these constants could encapsulate all the information that the below maps contain. Something along the line
```
MsgTypeFenceRequest = messageMeta("fence-req", Request, &proto.FenceRequest{})
```
We could still build a map, but then we didn't have to change this file in so many places whenever we add a new RPC.

maelstrom/main.go
merlimat: There is not client lib in this model, so it would need to handle the hashing of the keys too in the proxy layer.

maelstrom/messages.go
merlimat: There are a lot of improvements that could be done in this code, but this is already working, and it uncovered 4 bugs and gave a lot of visibility on what is happening inside, so I'd prefer to get this as it is. 

Other improvements can be added later on.

PR #170, opened 2022-12-29
Client lib should retry to perform requests within the configured timeout

merlimat:
Retry write and read batch requests multiple times, with exponential backoff, up to the max configured timeout.

Moved all the gRPC custom error codes in `common` module, so they are easily accessible for client lib to check for retriable errors.

### DIFF
--- a/cmd/client/cmd.go
+++ b/cmd/client/cmd.go
@@ -25,7 +25,7 @@ func init() {
 	Cmd.PersistentFlags().StringVarP(&common.Config.ServiceAddr, "service-address", "a", defaultServiceAddress, "Service address")
 	Cmd.PersistentFlags().DurationVar(&common.Config.BatchLinger, "batch-linger", oxia.DefaultBatchLinger, "Max time requests will be staged to be included in a batch")
 	Cmd.PersistentFlags().IntVar(&common.Config.MaxRequestsPerBatch, "max-requests-per-batch", oxia.DefaultMaxRequestsPerBatch, "Maximum requests per batch")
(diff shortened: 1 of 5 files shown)


cmd/client/cmd.go
```diff
@@ -25,7 +25,7 @@ func init() {
 	Cmd.PersistentFlags().StringVarP(&common.Config.ServiceAddr, "service-address", "a", defaultServiceAddress, "Service address")
 	Cmd.PersistentFlags().DurationVar(&common.Config.BatchLinger, "batch-linger", oxia.DefaultBatchLinger, "Max time requests will be staged to be included in a batch")
 	Cmd.PersistentFlags().IntVar(&common.Config.MaxRequestsPerBatch, "max-requests-per-batch", oxia.DefaultMaxRequestsPerBatch, "Maximum requests per batch")
-	Cmd.PersistentFlags().DurationVar(&common.Config.BatchRequestTimeout, "batch-request-timeout", oxia.DefaultBatchRequestTimeout, "Batch timeout in seconds")
+	Cmd.PersistentFlags().DurationVar(&common.Config.RequestTimeout, "batch-request-timeout", oxia.DefaultRequestTimeout, "Batch timeout in seconds")
```
nahguam:
```suggestion
	Cmd.PersistentFlags().DurationVar(&common.Config.RequestTimeout, "request-timeout", oxia.DefaultRequestTimeout, "Batch timeout in seconds")
```

nahguam: Should we remove the word `batch`? From the user's perspective, it could be easy to assume this is the timeout per request (put/get/delete etc)

merlimat: Yes, I think I changed it in few places, though not everywhere

PR #173, opened 2022-12-29
Allow to trim the head of the WAL

merlimat:
We need to be able to trim the WAL to expire old entries that are already in the DB.

### DIFF
--- a/server/wal/persistent_wal.go
+++ b/server/wal/persistent_wal.go
@@ -164,6 +164,7 @@ func (t *persistentWal) Clear() error {
 	}
 
 	t.lastOffset = InvalidOffset
+	t.firstOffset = InvalidOffset
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


server/wal/wal.go
```diff
@@ -44,8 +44,13 @@ type Wal interface {
 	io.Closer
 	// Append writes an entry to the end of the log
 	Append(entry *proto.LogEntry) error
+
+	// Trim removes all the entries that are
```
andrasbeni:
```suggestion
	// Trim removes all the entries that are before firstOffset
```

server/wal/persistent_wal.go
```diff
@@ -57,21 +58,24 @@ func newPersistentWal(shard uint32, options *WalFactoryOptions) (Wal, error) {
 		return nil, err
 	}
 
-	var lastOffset int64
+	var firstOffset, lastOffset int64
```
andrasbeni, line 61: I think `firstOffset` needs to be updated on truncation when it removes all entries.

merlimat: 👍  Fixed

PR #174, opened 2022-12-30
Improve the follower shutdown process

merlimat:
There are some intermittent tests failures (eg: https://github.com/streamnative/oxia/actions/runs/3807747148/jobs/6477719958 ) where the `server.Close()` operation is stuck and the test times out. 

From the stack traces, it looks like the follower close doesn't terminate the go-routine that handles the `AddEntries()` stream and that leaves the Grpc shutdown hanging.

Adding extra cancel function to make sure we always tear it down immediately.


server/follower_controller.go
```diff
@@ -64,21 +65,25 @@ type followerController struct {
 	wal         wal.Wal
 	kvFactory   kv.KVFactory
 	db          kv.DB
-	closeCh     chan error
-	log         zerolog.Logger
+
+	ctx           context.Context
+	cancel        context.CancelFunc
+	closeStreamCh chan error
```
andrasbeni, line 73: It looks to me that we don't need this channel anymore if we create the context in AddEntries/SendSnapshot. Am I missing something?

merlimat: This is needed to check the state of the currently connected leader stream and check the error in handling it, either the addEntries or the sendSnapshot stream

server/follower_controller.go
```diff
@@ -423,20 +438,26 @@ func (fc *followerController) SendSnapshot(stream proto.OxiaLogReplication_SendS
...
 
-	if fc.closeCh != nil {
+	if fc.closeStreamCh != nil {
 		fc.Unlock()
 		return ErrorLeaderAlreadyConnected
 	}
 
-	fc.closeCh = make(chan error)
+	closeStreamCh := make(chan error)
+	fc.closeStreamCh = closeStreamCh
```
nahguam, line 458: Is there any significance to having 2 vars?

server/follower_controller.go
```diff
@@ -148,10 +153,14 @@ func (fc *followerController) closeChannelNoMutex(err error) {
...
-	if fc.closeCh != nil {
-		fc.closeCh <- err
-		close(fc.closeCh)
-		fc.closeCh = nil
+	if fc.closeStreamCh != nil {
+		select {
+		case fc.closeStreamCh <- err:
+		default:
+			// Only write if there's a listener
+		}
```
nahguam, line 163: I'm finding this block a bit confusing. If there's nothing reading from the channel when the `select` is invoked, it will go through the `default` branch, correct?

It behaves differently if the `chan` is buffered - it will put the error on the channel regardless. I suppose that's because internally, there is something reading from the channel and buffering the error. We'll have to be careful not to make it buffered in the future!

server/follower_controller.go
merlimat: The `fc.closeStreamCh` can be modified by other go-routines and we need to read from the channel without holding the lock. 

Using the local variable fixes the data race.

server/follower_controller.go
merlimat: Yes, this is just to avoid getting blocked here in case the main go-routing handling the stream is already gone. 

It wouldn't change much if the channel is buffered: we wouldn't get blocked anyway. 

> It behaves differently if the chan is buffered - it will put the error on the channel regardless. 

That's ok for this usage, if the stream is already closed, we don't care of reporting an error back to the client anymore.

PR #176, opened 2022-12-30
Leader sends snapshot to followers when needed

merlimat:
The leader will send a snapshot to followers when either: 
 * The follower is getting bootstrapped
 * The follower is tailing at an entry that was already trimmed in the leader wal

(Stacked on top of #173)


server/follower_controller.go
```diff
@@ -469,15 +474,22 @@ func (fc *followerController) handleSnapshot(stream proto.OxiaLogReplication_Sen
...
+				break
+			}
 			fc.closeChannelNoMutex(err)
 			return
 		} else if snapChunk == nil {
 			break
-		} else if snapChunk.Epoch != fc.epoch {
+		} else if fc.epoch != wal.InvalidEpoch && snapChunk.Epoch != fc.epoch {
+			// The follower could be left with epoch=-1 by a previous failed
+			// attempt at sending the snapshot. It's ok to proceed int that case.
```
nahguam:
```suggestion
			// attempt at sending the snapshot. It's ok to proceed in that case.
```

server/follower_controller.go
```diff
@@ -500,11 +512,31 @@ func (fc *followerController) handleSnapshot(stream proto.OxiaLogReplication_Sen
 		return
 	}
 
+	// The new epoch must be persisted, to avoid rolling it back
+	if err = newDb.UpdateEpoch(fc.epoch); err != nil {
+		fc.closeChannelNoMutex(errors.Wrap(err, "failed to update epoch in db"))
```
nahguam:
```suggestion
		fc.closeChannelNoMutex(errors.Wrap(err, "Failed to update epoch in db"))
```

(or the 2 instances below to lowercase)

server/wal/wal.go
```diff
@@ -44,8 +44,13 @@ type Wal interface {
 	io.Closer
 	// Append writes an entry to the end of the log
 	Append(entry *proto.LogEntry) error
+
+	// Trim removes all the entries that are
```
nahguam: Looks like there should be more to the comment?

server/wal/wal_test.go
```diff
@@ -305,3 +306,66 @@ func Clear(t *testing.T) {
 	assert.NoError(t, w.Close())
 	assert.NoError(t, f.Close())
 }
+
+func Trim(t *testing.T) {
```
nahguam:
```suggestion
func TestTrim(t *testing.T) {
```

merlimat: This is called from a main test method, instead of being a top-level test.

nahguam: ah ok

PR #178, opened 2022-12-31
Trim the WAL after retention period expires

merlimat:
Introduced a wal trimmer component that does a binary search to identify the first non-expired offset where we need to periodically trim the WAL.

Defaults: 
 * Check every 10 minutes
 * Time retention of 1h in the WAL

This is stacked on top of #173


server/follower_controller.go
```diff
@@ -82,16 +84,15 @@ func NewFollowerController(shardId uint32, wf wal.WalFactory, kvFactory kv.KVFac
...
+	if fc.wal, err = wf.NewWal(shardId); err != nil {
 		return nil, err
-	} else {
-		fc.wal = wal
 	}
 
-	if db, err := kv.NewDB(shardId, kvFactory); err != nil {
+	fc.walTrimmer = wal.NewTrimmer(shardId, fc.wal, config.WalRetentionTime, wal.DefaultCheckInterval, wal.SystemClock)
+
+	if fc.db, err = kv.NewDB(shardId, kvFactory); err != nil {
```
andrasbeni, line 99: I think we should handle the case when opening the DB fails and the wal remains open. But not necessarily in this PR.

server/wal/wal_trimmer.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (t *trimmer) Close() error {
+	t.cancel()
+	t.ticker.Stop()
+
+	<-t.waitClose
+	return nil
+}
+
+func (t *trimmer) start() {
```
andrasbeni: I'd prefer a name that reflects that this method runs throughout the whole lifecycle of the object. Maybe `run()` or `waitForTicks()`.

server/wal/wal_trimmer.go
```diff
@@ -0,0 +1,178 @@
...
+		Time("timestamp-first-entry", tsFirst).
+		Time("cutoff-time", cutoffTime).
+		Msg("Starting wal trimming")
+
+	if cutoffTime.Before(tsFirst) {
+		// First entry has not expired. We don't need to check more
+		return nil
+	}
+
+	trimOffset, err := t.binarySearch(t.wal.FirstOffset(), t.wal.LastOffset(), cutoffTime)
```
andrasbeni, line 113: What would you say is the characteristic size of data retained with the default retention time? I'm asking because we have 20MB segments (the default is never overridden), and we cache two segments at any given time (also configurable). So if we have at least 3 segments, the binary search may jump between segments and thus, in the first few steps, opens and reads segments into memory just to evict them in the next step. 
I'd propose to iterate over segments and do a binary search only within segments if their last entry is to be retained. I know this is not possible through the current interface, but as this is the same package, we can access the WAL in ways we want to keep private.

server/wal/wal_trimmer_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for i := int64(0); i < 100; i++ {
+		assert.NoError(t, w.Append(&proto.LogEntry{
+			Epoch:     0,
+			Offset:    i,
+			Value:     []byte(""),
+			Timestamp: uint64(i),
+		}))
+	}
+
+	trimmer := NewTrimmer(1, w, 2*time.Millisecond, 10*time.Millisecond, clock)
```
andrasbeni, line 41: It's strange that you use a larger check interval than the retention time.

server/wal/wal_trimmer_test.go
```diff
@@ -0,0 +1,72 @@
...
+	assert.Eventually(t, func() bool {
+		log.Info().Int64("first-offset", w.FirstOffset()).Msg("checking...")
+		return w.FirstOffset() == 87
+	}, 10*time.Second, 10*time.Millisecond)
+
+	assert.NoError(t, trimmer.Close())
+
+	clock.currentTime.Store(95)
+
+	// The trimmer should be operating anymore once it's closed
```
andrasbeni:
```suggestion
	// The trimmer should not be operating anymore once it's closed
```

server/wal/wal_trimmer_test.go
merlimat: The 2 milliseconds is in "mocked" time, while the 10 are the interval for checking which is not mocked.

server/wal/wal_trimmer.go
merlimat: I think the current caching logic of the wal makes little sense as it is, at least for any use case that we are using the wal for.

There's no point in caching everything in the wal. We should just be caching the latest entries, as non-serialized objects for all the cursors to read.

And, we should have a separate index for the wal that will allow to quickly find the position on disk for a particular wal offset. 

The index is simply an array backed by a file: `[]uint32` : `{firstFileOffset, secondFileOffset, ...}`.. 
Checking for the physical location of a logical offset is an O(1) operation: 
 1. Check the index file at a specific offset
 2. Read in the wal file at the location.

No cache is needed here.

Additionally, the wal segment files sizes should be increased in size, in a range of 256MB - 2GB

andrasbeni: Created the backlog item https://www.notion.so/streamnativeio/Change-WAL-caching-logic-9c6e8bfa7db0418eb3b2bdc96f6131b8

server/wal/wal_trimmer_test.go
andrasbeni: I know. It's still strange to give parameters this way.

server/wal/wal_trimmer.go
```diff
@@ -0,0 +1,178 @@
...
+const (
+	DefaultRetention     = 1 * time.Hour
+	DefaultCheckInterval = 10 * time.Minute
+)
+
+type Trimmer interface {
+	io.Closer
+}
+
+type Clock interface {
```
nahguam: We have a clock interface in common, albeit its signature is:

```
type Clock interface {
	NowMillis() uint64
}
```

It's only used in 1 place so far, so we could refactor it to the signature defined here.

server/wal/wal_trimmer.go
nahguam: +1 `run()`

server/wal/wal_trimmer.go
merlimat: Can we standardize on returning `time.Time`?

server/follower_controller.go
merlimat: Good point. I started looking at it, though it's going to need a more thoughtful approach, because there are so many different exit points in the method. (and many other similar cases that we should address). 

Perhaps we should standardize in calling `fc.Close()` and make sure the close method can avoid closing the `nil` pointers.

server/wal/wal_trimmer.go
nahguam: 👍

merlimat: Fixed

PR #179, opened 2023-01-03
Adding metrics wrapper interfaces and WAL metrics

merlimat:
Added wrapper interface on top of OpenTelemetry and added metrics for WAL.

### DIFF
--- a/server/wal/persistent_wal.go
+++ b/server/wal/persistent_wal.go
@@ -51,6 +51,8 @@ type persistentWal struct {
 	readLatency   metrics.LatencyHistogram
 	readBytes     metrics.Counter
 	trimOps       metrics.Counter
+	readErrors    metrics.Counter
+	writeErrors   metrics.Counter
 	activeEntries metrics.Gauge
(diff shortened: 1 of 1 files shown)


server/wal/persistent_wal.go
```diff
@@ -58,29 +67,47 @@ func newPersistentWal(shard uint32, options *WalFactoryOptions) (Wal, error) {
...
-		lastOffset:  lastOffset,
-		firstOffset: firstOffset,
+		w.lastOffset = lastEntry.Offset
+		w.firstOffset = log.firstOffset
 	}
 	return w, nil
 }
 
-func readAtIndex(log *Log, index int64) (*proto.LogEntry, error) {
+func (t *persistentWal) readAtIndex(log *Log, index int64) (*proto.LogEntry, error) {
```
andrasbeni: nit: we could use `t.log` instead of the parameter.

merlimat: 👍 I missed this

PR #180, opened 2023-01-04
Remove unnecessary block

andrasbeni:
Motivation: https://github.com/streamnative/oxia/pull/120/files#r1060781521

### DIFF
--- a/common/condition.go
+++ b/common/condition.go
@@ -49,13 +49,13 @@ type conditionContext struct {
 	sync.RWMutex
 	locker sync.Locker
 
-	ch chan any
+	ch chan bool
 }
 
 func NewConditionContext(locker sync.Locker) ConditionContext {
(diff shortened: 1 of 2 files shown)


server/leader_controller.go
```diff
@@ -416,12 +416,10 @@ func (lc *leaderController) Read(request *proto.ReadRequest) (*proto.ReadRespons
...
 		Msg("Received read request")
 
-	{
-		lc.Lock()
-		defer lc.Unlock()
-		if err := checkStatus(proto.ServingStatus_Leader, lc.status); err != nil {
-			return nil, err
-		}
+	lc.Lock()
+	defer lc.Unlock()
```
merlimat: We should unlock before the `ProcessRead()`. Otherwise we have a single read/write operation at a time.

PR #182, opened 2023-01-04
Fixed notifications API

merlimat:
There is a flaky test for the notification client API (https://github.com/streamnative/oxia/actions/runs/3840904899/jobs/6540510030 ) that is uncovering a problem with the API itself. 

The current API is 

```go
type Client interface {
     // ... 
     GetNotifications() <-chan Notification
}
```

The idea is that when an application calls this, it will start receiving notifications from the "next transaction." 

The problem is that we are returning the channel and setting up the stream while the gRPC streams (one per shard) are being set up in the background. This could lead to: 

```
ch := client.GetNotifications()
client.Put("my-key", "my-value")

notification := <-ch // We could miss the notification for the previous transaction.
```

We must ensure that after `GetNotifications()` returns, we don't miss any new transactions. To achieve that, we need the request to go through the server. With the gRPC streams, there is no good way to achieve it other than returning a first dummy notification that will carry the last committed offset.

### Modifications 
 1. Changed the API into : 
```go
type Client interface {
     // ... 
     GetNotifications() (Notifications, error)
}

type Notifications interface {
	io.Closer

	Ch() <-chan *Notification
}
```
 2. Wait until we have initialized the notifications on all the shards before returning the `Notifications`
 3. Returning `Notifications` instead of the bare `chan Notification` to allow an application to close it and unsubscribe

### DIFF
--- a/oxia/notifications.go
+++ b/oxia/notifications.go
@@ -13,7 +13,7 @@ import (
 	"time"
 )
 
-type notificationsManager struct {
+type notifications struct {
 	multiplexCh  chan *Notification
 	closeCh      chan any
 	shardManager internal.ShardManager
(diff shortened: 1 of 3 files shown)


oxia/notifications.go
```diff
@@ -144,6 +179,18 @@ func (snm *shardNotificationsManager) getNotifications() error {
...
 			Msg("Received batch notification")
 
+		if !snm.initialized {
+			snm.log.Debug().Msg("Initialized the notification manager")
+
+			// We need to discard the very first notification, because it's only
+			// needed to ensure that the notification cursor is created on the
+			// server side.
+			snm.initialized = true
+			snm.nm.initialized <- nil
```
nahguam, line 189: nit: Would the intent be clearer by closing the channel, instead of sending `nil`?

merlimat: We need to wait until all the shards have been initiated or until the first failure. 

Thinking about it, we could encapsulate this logic, which we're also needing in other places, into a sort of "WaitGroup with a ctx and an error returned "

PR #184, opened 2023-01-05
Extract stream reader

andrasbeni:
The stream reader abstraction is based on the follower controller's handling of AddEntries.
This abstraction will be reused in session management.

### DIFF
--- a/server/assignment_dispatcher.go
+++ b/server/assignment_dispatcher.go
@@ -9,6 +9,7 @@ import (
 	"google.golang.org/grpc/status"
 	"io"
 	"math"
+	"oxia/common"
 	"oxia/proto"
 	"oxia/server/util"
 	"sync"


server/assignment_dispatcher.go
```diff
@@ -109,14 +109,23 @@ func (s *shardAssignmentDispatcher) Initialized() bool {
 	return s.assignments != nil
 }
 
+func (s *shardAssignmentDispatcher) logError(err error) {
```
merlimat: I think it would be simpler to just pass the `s.log` into the `streamReader`

server/assignment_dispatcher.go
```diff
@@ -109,14 +109,23 @@ func (s *shardAssignmentDispatcher) Initialized() bool {
...
+		stream,
+		s.updateShardAssignment,
+		s.logError,
+		map[string]string{
+			"oxia": "receive-shards-assignments",
+		},
+	)
 	select {
-	case err := <-ch:
+	case err := <-streamReader.Run():
```
merlimat: I would actually pass the `s.ctx` and move the select block into the streamReader and just do 

```go
return streamReader.Run()
```

server/util/stream_reader.go
```diff
@@ -0,0 +1,84 @@
...
+		} else if !isSender {
+			continue
+		} else if err = sender.Send(res); err != nil {
+			s.Close(err)
+			return
+		}
+	}
+}
+
+func (s *streamReader[T, U]) Close(err error) {
```
merlimat: Shouldn't this be private? It's modifying the `s.closeCh` without a mutex, so it wouldn't be safe to be used from other go-routines

server/util/stream_reader.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if s.closeCh != nil {
+		s.closeCh <- err
+		close(s.closeCh)
+		s.closeCh = nil
+	}
+}
+
+func ReadStream[T any, U any](
```
merlimat: We could have a wrapper for mono-directional streams, where the `handleMessage` is just `func(*T) error`

PR #195, opened 2023-01-09
Added CLI command for printing notifications

merlimat:


cmd/client/notifications/cmd.go
```diff
@@ -0,0 +1,41 @@
...
+	RunE:  exec,
+}
+
+func exec(cmd *cobra.Command, args []string) error {
+	client, err := common.Config.NewClient()
+	if err != nil {
+		return err
+	}
+
+	defer client.Close()
```
nahguam, line 23: This swallows errors (and below)? Do we care?

cmd/client/notifications/cmd.go
```diff
@@ -0,0 +1,41 @@
...
+
+	notifications, err := client.GetNotifications()
+	if err != nil {
+		return err
+	}
+
+	defer notifications.Close()
+
+	for notification := range notifications.Ch() {
```
nahguam, line 32: Does this clean close on interrupt?

merlimat: yes, it exits on Ctrl-C

cmd/client/notifications/cmd.go
merlimat: I think it should be ok. It just runs and receives notifications until it's stopped

PR #199, opened 2023-01-10
Add basic operator controller

nahguam:


deploy/crds/oxiaclusters.yaml
```diff
@@ -17,72 +17,96 @@ spec:
...
+                initialShardCount:
                   type: integer
+                  x-kubernetes-validations:
+                    - message: InitialShardCount is immutable
+                      rule: self == oldSelf
                 replicationFactor:
                   type: integer
+                  x-kubernetes-validations:
+                    - message: ReplicationFactor is immutable
+                      rule: self == oldSelf
```
merlimat, line 35: Could we set here `1 <= replicationFactor <= 15` ?

nahguam: We can - I have a TODO and ticket to implement validations for all the fields. I'll do that in a follow up tomorrow

PR #202, opened 2023-01-11
Added grafana dashboards

merlimat:
Added a set of grafana dashboards: 

 * Containers -> Container cpu/mem/disk
 * Coordinator ->  Leader elections and nodes health checks
 * Golang -> Go process metrics
 * Grpc -> client & server request rates
 * Nodes -> Per server node aggregated view of db metrics
 * Shards -> Per shard view of db metrics

### DIFF
--- a/deploy/dashboards/oxia-containers.json
+++ b/deploy/dashboards/oxia-containers.json
@@ -24,7 +24,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 33,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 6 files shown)


deploy/dashboards/oxia-containers.json
```diff
@@ -0,0 +1,782 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "$$hashKey": "object:4632",
+        "builtIn": 1,
+        "datasource": {
+          "uid": "${DataSource}"
+        },
```
nahguam, line 9: Some have variables of `${DataSource}` like here, and some are hardcoded to the prometheus datasource:

```
      "datasource": {
        "type": "prometheus",
        "uid": "prometheus"
      },
```

There is also no `${DataSource}` variable defined resulting in an error.

We should make them all use the `${DataSource}` variable and define it.

nahguam: I've done this

merlimat: Good catch. I had imported (and then modified) a couple of dashboards

PR #208, opened 2023-01-13
Client side session management

andrasbeni:


oxia/async_client_impl_test.go
```diff
@@ -25,27 +25,27 @@ func TestAsyncClientImpl(t *testing.T) {
...
 		Stat:    putResult.Stat,
 	}, getResult)
 
-	putResult = <-client.Put("/c", []byte{0}, &VersionNotExists)
-	assert.Equal(t, versionZero, putResult.Stat.Version)
+	putResult = <-client.Put("/c", []byte{0}, ExpectedVersion(VersionNotExists))
+	assert.Equal(t, int64(0), putResult.Stat.Version)
 
-	putResult = <-client.Put("/c", []byte{1}, &versionZero)
+	putResult = <-client.Put("/c", []byte{1}, ExpectedVersion(versionZero))
```
merlimat: I think the only reason to have `versionZero` here was that we couldn't take a pointer to 0 (`&0`). 


```suggestion
	putResult = <-client.Put("/c", []byte{1}, ExpectedVersion(0))
```

oxia/async_client_impl_test.go
```diff
@@ -25,27 +25,27 @@ func TestAsyncClientImpl(t *testing.T) {
 	client, err := NewAsyncClient(serviceAddress, WithBatchLinger(0))
 	assert.NoError(t, err)
 
-	putResult := <-client.Put("/a", []byte{0}, &VersionNotExists)
-	assert.Equal(t, versionZero, putResult.Stat.Version)
+	putResult := <-client.Put("/a", []byte{0}, ExpectedVersion(VersionNotExists))
+	assert.Equal(t, int64(0), putResult.Stat.Version)
```
merlimat: nit: 

```suggestion
	assert.EqualValues(t, 0, putResult.Stat.Version)
```

oxia/options.go
```diff
@@ -14,13 +14,16 @@ const (
 	DefaultBatchLinger         = 5 * time.Millisecond
 	DefaultMaxRequestsPerBatch = 1000
 	DefaultRequestTimeout      = 30 * time.Second
+	DefaultSessionTimeout      = 5 * time.Minute
```
merlimat: I'd leave 5min as the "max" session timeout, though for the default value, we should have something around 10/30 seconds.

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc/metadata"
+	"oxia/common"
+	"oxia/oxia/internal"
+	"oxia/proto"
+	"sync"
+	"time"
+)
+
+func NewSessions(ctx context.Context, shardManager internal.ShardManager, pool common.ClientPool, options clientOptions) *sessions {
```
merlimat: This shouldn't be part of public API:
```suggestion
func newSessions(ctx context.Context, shardManager internal.ShardManager, pool common.ClientPool, options clientOptions) *sessions {
```

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (s *sessions) executeWithSessionId(ctx context.Context, shardId uint32, callback func(int64, error)) {
+	s.Lock()
+	session, found := s.sessionsByShard[shardId]
+	if !found {
+		session = s.startSession(ctx, shardId)
+		s.sessionsByShard[shardId] = session
+	}
+	session.executeWithId(ctx, callback)
+	defer s.Unlock()
```
merlimat: Move `defer s.Unlock()` to 2nd line in method (or remove `defer`)

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+
+}
+
+func (s *sessions) startSession(ctx context.Context, shardId uint32) *clientSession {
+	cs := &clientSession{
+		shardId:  shardId,
+		sessions: s,
+		log: log.With().
+			Str("component", "session").
+			Uint32("shard-id", shardId).Logger(),
```
merlimat: We've been using `shard` elsewhere

```suggestion
			Uint32("shard", shardId).Logger(),
```

oxia/options.go
merlimat: We could also define a min-session-timeout setting of 2-5 seconds.

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+		"oxia":  "session-keep-alive",
+		"shard": fmt.Sprintf("%d", cs.shardId),
+	}, func() { cs.keepAlive(rpc) })
+
+	return nil
+}
+
+func (cs *clientSession) keepAlive(rpc proto.OxiaClientClient) {
+	cs.sessions.Lock()
+	cs.Lock()
```
merlimat, line 197: We need to be careful with the ordering of these locks :)

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+	timeout := cs.sessions.clientOpts.sessionTimeout
+	ctx := cs.sessions.ctx
+	shardId := cs.shardId
+	sessionId := cs.sessionId
+	cs.Unlock()
+	cs.sessions.Unlock()
+
+	timer := time.NewTicker(2 * timeout / 3)
```
merlimat: We shouldn't be waiting until 2/3 of session timeout to send the heath-check. Otherwise a small network hiccup could cause the session to expire. 

eg: 
 * `sessionTimeout=12sec`
 * We wait for 8sec before sending keep-alive
 * There's a 5sec network congestion
 * The session expires

We should instead take the interval to something like `max(sessionTimeout/10, 2sec)`

This way we can survive a longer network partititon.

oxia/sessions.go
```diff
@@ -0,0 +1,188 @@
...
+			err = client.Send(&proto.SessionHeartbeat{
+				ShardId:   shardId,
+				SessionId: sessionId,
+			})
+			if err != nil {
+				// TODO log, retry
+				return
+			}
+		case <-ctx.Done():
+			// TODO rpc.Close()?
```
merlimat: Yes, if we're closing the client, we should force the session to be closed, so that all the ephemerals are immediately released.

oxia/sessions.go
merlimat: Looks correct so far, though I was having shiverings :D

andrasbeni: I try to be careful. Thanks for double-checking.

oxia/sessions.go
andrasbeni: Makes sense.

oxia/sessions.go
andrasbeni: My problem is that the ctx is already "done" so it cannot be used to initiate an rpc. Do you think I should use a new context?

oxia/sessions.go
```diff
@@ -0,0 +1,202 @@
...
+	sessionId := cs.sessionId
+	cs.Unlock()
+	cs.sessions.Unlock()
+
+	tickTime := timeout / 10
+	if tickTime < 2*time.Second {
+		tickTime = 2 * time.Second
+	}
+
+	timer := time.NewTicker(tickTime)
```
merlimat: not sure it's 100% required, though we could have here 
`defer timer.Stop()`

oxia/sessions.go
```diff
@@ -0,0 +1,202 @@
...
+			if err != nil {
+				// TODO log, retry
+				return
+			}
+		case <-ctx.Done():
+			err = client.CloseSend()
+			if err != nil {
+				cs.log.Error().Err(err).Msg("Failed to close heartbeat stream.")
+			}
+			_, err = rpc.CloseSession(ctx, &proto.CloseSessionRequest{
```
merlimat, line 245: If the `ctx` is already done, this will fail immediately. We should have a new `ctx`, using the default timeout.

oxia/sessions.go
```diff
@@ -0,0 +1,202 @@
...
+	for {
+		select {
+		case <-timer.C:
+			err = client.Send(&proto.SessionHeartbeat{
+				ShardId:   shardId,
+				SessionId: sessionId,
+			})
+			if err != nil {
+				// TODO log, retry
+				return
```
merlimat: We should just return `err` here and add the `backoff.Retry()` as we do in other places.

PR #213, opened 2023-01-16
minor controller cleanup

nahguam:


controller/watcher.go
```diff
@@ -22,10 +22,6 @@ import (
 	oxia "oxia/pkg/generated/clientset/versioned"
 )
 
-type Watcher interface {
```
andrasbeni, line 25: I believe using the type name `Watcher` adds to readability because `Closer` is too generic a concept and helps navigate in the code. Would you mind keeping it as an alias?
```
type Watcher io.Closer
```

nahguam: 👍

PR #217, opened 2023-01-17
Fix and improve controller & tests

nahguam:
* mock health client immediately send status on watch start
* inject reduced backoff to allow tests to be faster


coordinator/impl/coordinator_e2e_test.go
```diff
@@ -181,7 +181,7 @@ func TestCoordinator_LeaderFailover(t *testing.T) {
 	assert.NoError(t, client.Close())
 
 	// Stop the leader to cause a leader election
-	assert.NoError(t, servers[leader].Close())
+	assert.NoError(t, servers[leader].Close()) //TODO blocks here
```
nahguam: There appears to be a race somewhere. If I just run the test it blocks here. If I step through with the debugger, it passes.

PR #218, opened 2023-01-17
Break Snapshot chunks into small gRPC requests

andrasbeni:


server/kv/kv_pebble.go
```diff
@@ -622,7 +688,46 @@ func (sl *pebbleSnapshotLoader) Close() error {
 }
 
 func (sl *pebbleSnapshotLoader) AddChunk(name string, content []byte) error {
-	return os.WriteFile(filepath.Join(sl.dbPath, name), content, 0644)
+	r := regexp.MustCompile(`^(.+):(\d+)/(\d+)$`)
```
merlimat: If we're getting multiple chunks with the same name, couldn't we just append to the existing file?

server/kv/kv_pebble.go
```diff
@@ -570,28 +580,84 @@ func (ps *pebbleSnapshot) Valid() bool {
...
 
-func (ps *pebbleSnapshot) Chunk() SnapshotChunk {
-	return &pebbleSnapshotChunk{ps, ps.files[0]}
+func (ps *pebbleSnapshot) Chunk() (SnapshotChunk, error) {
+	content, err := ps.NextChunkContent()
+	if err != nil {
+		return nil, err
+	}
+	return &pebbleSnapshotChunk{
+		fmt.Sprintf("%s:%d/%d", ps.files[0], ps.chunkIndex, ps.chunkCount),
```
merlimat: Instead of having informal definition in the name, we could add fields in the proto structure

server/kv/kv_pebble.go
andrasbeni: It would be possible, just like I do it in `TestPebbleSnapshot`. It's simpler, but I wanted to avoid the overhead of closing and reopening the file unnecessarily (which would have to happen if I didn't know which are the last chunks of a file). If you say that overhead is negligible, I'm all for simplifying the logic.

andrasbeni: As discussed in the meeting today, I'll proceed by keeping the current logic but separating the progress numbers into their own fields in the protocol instead of encoding them as part of the name.

server/kv/kv_pebble.go
```diff
@@ -570,28 +580,94 @@ func (ps *pebbleSnapshot) Valid() bool {
...
-func (ps *pebbleSnapshot) Chunk() SnapshotChunk {
-	return &pebbleSnapshotChunk{ps, ps.files[0]}
+func (ps *pebbleSnapshot) Chunk() (SnapshotChunk, error) {
+	content, err := ps.NextChunkContent()
+	if err != nil {
+		return nil, err
+	}
+	return &pebbleSnapshotChunk{
+		ps.files[0],
+		int32(ps.chunkIndex),
```
merlimat, line 638: I'd keep the same type in both structs

server/kv/kv.go
```diff
@@ -67,14 +71,14 @@ type Snapshot interface {
...
-	Chunk() SnapshotChunk
+	Chunk() (SnapshotChunk, error)
 	Next() bool
 }
 
 type SnapshotLoader interface {
 	io.Closer
 
-	AddChunk(name string, content []byte) error
+	AddChunk(string, int32, int32, []byte) error
```
merlimat: Let's leave the argument Mamés in the interface, otherwise it becomes hard to understand what they mean

PR #219, opened 2023-01-17
Checkin generated code

nahguam:


proto/client.pb.go
```diff
@@ -0,0 +1,2628 @@
+// Copyright 2023 StreamNative, Inc.
```
merlimat, line 1: We should probably just ignore the license header on these generated files, otherwise we'd have to manually fix them each time we regenerate them.

In the `license-check`/`license-format` we can filter for `-v '*.pb.go'`

nahguam: The license rules do indeed check the generated code. However, `protoc-gen-go` already copies the license header over from the proto files. `proto-gen-go-grpc` does not do this in v1.2.0 so `license-format` is adding them, but the next release will as this feature has now been [added](https://github.com/grpc/grpc-go/pull/5540).

So the net result is the same - the generated `*.pb.go` files will have the license header in them, so I don't think it does any harm to continue to verify they are there.

PR #221, opened 2023-01-18
Trim DB notifications after retention time expires

merlimat:


server/kv/notifications_trimmer_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"context"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"oxia/proto"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+type mockedClock struct {
```
andrasbeni: Can't we just reuse `mockClock` from [wal_trimmer_test.go](https://github.com/streamnative/oxia/blob/main/server/wal/wal_trimmer_test.go#L31-L37)?

merlimat: Good point. Fixed

PR #232, opened 2023-01-19
Renamed Fence -> NewTerm

merlimat:


coordinator/impl/shard_controller.go
```diff
@@ -295,13 +295,13 @@ func (s *shardController) keepFencingFollower(ctx context.Context, node model.Se
 				Interface("follower", node).
 				Int64("term", s.Term()).
 				Dur("retry-after", duration).
-				Msg("Failed to fence, retrying later")
+				Msg("Failed to newTerm, retrying later")
```
andrasbeni, line 298: nit: `newTerm` as a verb sounds unusual.

PR #257, opened 2023-02-07
Streaming read

nahguam:


server/leader_controller.go
```diff
@@ -516,19 +521,46 @@ func (lc *leaderController) truncateFollowerIfNeeded(follower string, followerHe
...
+func (lc *leaderController) read(ctx context.Context, request *proto.ReadRequest, ch chan<- GetResult) {
+	common.DoWithLabels(map[string]string{
+		"oxia":  "read",
+		"shard": fmt.Sprintf("%d", lc.shardId),
+		"peer":  common.GetPeer(ctx),
+	}, func() {
+		lc.log.Debug().
+			Msg("Received read request")
+
+		for _, get := range request.Gets {
```
merlimat, line 551: Would this be sending 1 response for each item in the batched-read request? Could it have a performance penalty compared to sending back many responses together (within the max-frame size)?

nahguam: At this layer, it's sending them individually to the `chan`. The re-batching for the grpc response happens up a layer in `publicRpcServer.Read`

PR #259, opened 2023-02-13
perf benchmark

nahguam:


cmd/perf/cmd.go
```diff
@@ -17,43 +17,23 @@ package perf
...
-	MaxRequestsPerBatch int
-	RequestTimeout      time.Duration
-}
-
 var (
 	Cmd = &cobra.Command{
-		Use:   "perf",
-		Short: "Oxia perf client",
+		Use:   "runPerf",
+		Short: "Oxia runPerf client",
```
merlimat:
```suggestion
		Use:   "perf",
		Short: "Oxia perf client",
```

nahguam: oops, goland's rename feature is a bit overzealous sometimes

PR #269, opened 2023-03-01
Limit batch size

andrasbeni:


cmd/client/delete/cmd_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"testing"
 )
 
-func TestCorba(t *testing.T) {
+func TestCobra(t *testing.T) {
```
merlimat, line 26: `Corba` brought back some very bad memories :)

oxia/options.go
```diff
@@ -27,13 +27,15 @@ import (
 const (
 	DefaultBatchLinger         = 5 * time.Millisecond
 	DefaultMaxRequestsPerBatch = 1000
+	DefaultMaxBatchSize        = 4 * 1024 * 1024
```
merlimat: I'd start with a lower max limit. Probably after `128Kb` there's no practical improvement in throughput.

oxia/options.go
```diff
@@ -117,6 +121,18 @@ func WithMaxRequestsPerBatch(maxRequestsPerBatch int) ClientOption {
 	})
 }
 
+// WithMaxBatchSize defines an upper limit for the size of write batches.
+// The value must be greater than zero.
+func WithMaxBatchSize(maxBatchSize int) ClientOption {
```
merlimat: I'd even prefer to not let this be configurable. Rather just set the default max always.

PR #274, opened 2023-03-07
move GetHighestEntryOfTerm to leaderController

nahguam:
This func is only used in this one place, so it logically belongs here.


server/leader_controller.go
```diff
@@ -524,6 +524,27 @@ func (lc *leaderController) truncateFollowerIfNeeded(follower string, followerHe
 	}
 }
 
+func GetHighestEntryOfTerm(w wal.Wal, term int64) (*proto.EntryId, error) {
```
merlimat: If it's only used here, maybe we should convert to private:
```suggestion
func getHighestEntryOfTerm(w wal.Wal, term int64) (*proto.EntryId, error) {
```

PR #280, opened 2023-03-13
Wait for all nodes to be ready

nahguam:
Until we implement proper recovery, it is safer to wait for all nodes to be ready always, not just on first time start.

### DIFF
--- a/server/follower_controller.go
+++ b/server/follower_controller.go
@@ -125,7 +125,8 @@ func NewFollowerController(config Config, shardId uint32, wf wal.WalFactory, kvF
 	}
 
 	fc.lastAppendedOffset = fc.wal.LastOffset()
-	fc.walTrimmer = wal.NewTrimmer(shardId, fc.wal, config.WalRetentionTime, wal.DefaultCheckInterval, common.SystemClock)
+	fc.walTrimmer = wal.NewTrimmer(shardId, fc.wal, config.WalRetentionTime, wal.DefaultCheckInterval,
+		common.SystemClock, fc)
 
 	if fc.db, err = kv.NewDB(shardId, kvFactory, config.NotificationsRetentionTime, common.SystemClock); err != nil {
 		return nil, err
--- a/server/leader_controller.go
+++ b/server/leader_controller.go
@@ -142,7 +142,8 @@ func NewLeaderController(config Config, shardId uint32, rpcClient ReplicationRpc
 		return nil, err
 	}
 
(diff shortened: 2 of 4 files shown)


coordinator/impl/coordinator.go
```diff
@@ -95,13 +95,13 @@ func NewCoordinator(metadataProvider MetadataProvider, clusterConfig model.Clust
...
-		// Before initializing the cluster, it's better to make sure we
-		// have all the nodes available, otherwise the coordinator might be
-		// the first component in getting started and will print out a lot
-		// of error logs regarding failed leader elections
-		c.waitForAllNodesToBeAvailable()
+	// Before initializing the cluster, it's better to make sure we
+	// have all the nodes available, otherwise the coordinator might be
+	// the first component in getting started and will print out a lot
+	// of error logs regarding failed leader elections
+	c.waitForAllNodesToBeAvailable()
```
merlimat, line 102: This could lead to a case where 1 storage node is down, the coordinator restarts and won't be able to re-assign the leadership to other nodes, because of this wait

nahguam: Yes I know, but this is only on coordinator start up. My intent is that this is temporary until we implement better recovery. It's better than the current situation where it generates a bunch of errors due to no connection, then stalls somewhere - possibly related to the current errors we're looking at or something new.

Also worth noting that the coordinator also currently gets stuck if a server listed in the cluster-status is not accessible.

merlimat: >  It's better than the current situation where it generates a bunch of errors due to no connection, 

Error messages are not ideal, though they should settle when the backoff increases. Still I wouldn't trade it for the risk of not electing the leader on a shard for some extended amount of time, when a storage node is failing and the coordinator cannot fully initialize.

> then stalls somewhere - possibly related to the current errors we're looking at or something new.

the coordinator should get stalled -- If we have any place where that happens we need to fix them

> Also worth noting that the coordinator also currently gets stuck if a server listed in the cluster-status is not accessible.

The health of each server is checked from a different go-routine. Similarly, all the `NewTerm()` are done in parallel. Where does it get stuck?

nahguam: I agree. I don't know where it gets stuck, I haven't had time to check. As I say I intended this to be a temporary workaround until we can work through the problems. It's not essential. The other workaround is to wait for the pods to come online then restart the coordinator again. I just got a bit bored of doing that.

Happy to close this though

PR #281, opened 2023-03-14
Fixed trimming wal only up to commit offset

merlimat:
If the traffic stops on a shard, currently, the wal trimmer can trim all the entries in the log, even entries that are not yet fully committed. That will cause the leader to get stuck when pushing to a follower. 

The sequence is: 
 1. The follower is at an offset that is no longer available in the leader wal
 2. The leader sends a snapshot
 3. The snapshot is associated with an offset in the leader that no longer exists, so the streaming replication fails, and we go back to step (1).

```
{"level":"info","component":"follower-cursor","shard":1,"term":95,"follower":"oxia-2.oxia:6649","follower-ack-offset":4956,"wal-first-offset":4958,"wal-last-offset":4958,"time":"2023-03-13T14:04:20.664248691Z","message":"The follower is behind the first available entry in the leader WAL"}

{"level":"info","component":"follower-cursor","shard":1,"term":95,"follower":"oxia-2.oxia:6649","chunks-count":8,"total-size":"2.7 MiB","elapsed-time":"125.842627ms","throughput":"22 MiB/s","follower-ack-offset":4956,"time":"2023-03-13T14:04:24.781595964Z","message":"Successfully sent snapshot to follower"}

{"level":"error","component":"follower-cursor","shard":1,"term":95,"follower":"oxia-2.oxia:6649","stack":[{"func":"init","line":"25","source":"wal.go"},{"func":"doInit","line":"6329","source":"proc.go"},{"func":"doInit","line":"6306","source":"proc.go"},{"func":"doInit","line":"6306","source":"proc.go"},{"func":"doInit","line":"6306","source":"proc.go"},{"func":"main","line":"233","source":"proc.go"},{"func":"goexit","line":"1594","source":"asm_amd64.s"}],"error":"oxia: entry not found","retry-after":1482.562776,"time":"2023-03-13T14:04:24.781990978Z","message":"Error while pushing entries to follower"}

```

### DIFF
--- a/server/wal/wal_trimmer.go
+++ b/server/wal/wal_trimmer.go
@@ -136,7 +136,7 @@ func (t *trimmer) doTrim() error {
 		return errors.Wrap(err, "failed to perform binary search")
 	}
 
-	// We cannot trim past the commit offset, or we won't be able to
+	// We cannot trim past the commit offset, or we won't be able to replicate those entries
 	commitOffset := t.commitOffsetProvider.CommitOffset()
 	if commitOffset < trimOffset {
 		trimOffset = commitOffset


server/wal/wal_trimmer.go
```diff
@@ -129,6 +136,12 @@ func (t *trimmer) doTrim() error {
 		return errors.Wrap(err, "failed to perform binary search")
 	}
 
+	// We cannot trim past the commit offset, or we won't be able to
```
andrasbeni:
```suggestion
	// We cannot trim past the commit offset, or we won't be able to replicate those entries
```

PR #309, opened 2023-04-06
Set metadata.ownerReferences on cluster resources

coderzc:
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

### DIFF
--- a/kubernetes/resources.go
+++ b/kubernetes/resources.go
@@ -44,11 +44,26 @@ func resourceName(component Component, name string) string {
 	return name
 }
 
-func objectMeta(component Component, name string) metaV1.ObjectMeta {
+func objectMeta(component Component, name string, cluster *v1alpha1.OxiaCluster) metaV1.ObjectMeta {
 	_resourceName := resourceName(component, name)
-	return metaV1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


kubernetes/resources.go
```diff
@@ -194,12 +206,12 @@ func coordinatorDeployment(cluster v1alpha1.OxiaCluster) *appsV1.Deployment {
 		command = append(command, "--profile")
 	}
 	deployment := &appsV1.Deployment{
-		ObjectMeta: objectMeta(Coordinator, cluster.Name),
+		ObjectMeta: objectMeta(&cluster, Coordinator, cluster.Name),
 		Spec: appsV1.DeploymentSpec{
 			Replicas: pointer.Int32(1),
 			Selector: &metaV1.LabelSelector{MatchLabels: selectorLabels(Coordinator, cluster.Name)},
 			Template: coreV1.PodTemplateSpec{
-				ObjectMeta: objectMeta(Coordinator, cluster.Name),
```
mattisonchao, line 202: pod owner should be deployment?

coderzc: You are right, don't set metadata.ownerReferences for `PodTemplateSpec`.

PR #310, opened 2023-04-06
Support scrape metrics when using helm

mattisonchao:
## Motivation

Since some environments use the helm chart, but the Prometheus operator deploys the Prometheus cluster. That means the new change on `serviceMonitor` will not be reflected in Prometheus configuration `serviceMonitorSelector`. 

It's better to support the annotation method.

### DIFF
--- a/deploy/charts/oxia-cluster/templates/coordinator-deployment.yaml
+++ b/deploy/charts/oxia-cluster/templates/coordinator-deployment.yaml
@@ -26,8 +26,8 @@ spec:
   template:
     metadata:
       annotations:
-        prometheus.io/port: {{ .Values.coordinator.ports.metrics }}
-        prometheus.io/scrape: {{ .Values.monitoringEnabled }}
+        prometheus.io/port: "{{ .Values.coordinator.ports.metrics }}"
+        prometheus.io/scrape: "{{ .Values.monitoringEnabled }}"
       labels:
         {{- include "oxia-cluster.coordinator.labels" . | nindent 8 }}
       name: {{ .Release.Name }}-coordinator


deploy/charts/oxia-cluster/templates/coordinator-deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
       {{- include "oxia-cluster.coordinator.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      annotations:
+        prometheus.io/port: {{ .Values.coordinator.ports.metrics }}
+        prometheus.io/scrape: {{ .Values.monitoringEnabled }}
```
merlimat:
```suggestion
      annotations:
        prometheus.io/port: "{{ .Values.coordinator.ports.metrics }}"
        prometheus.io/scrape: "{{ .Values.monitoringEnabled }}"
```

deploy/charts/oxia-cluster/templates/server-statefulset.yaml
```diff
@@ -27,6 +27,9 @@ spec:
   podManagementPolicy: Parallel
   template:
     metadata:
+      annotations:
+        prometheus.io/port: {{ .Values.server.ports.metrics }}
+        prometheus.io/scrape: {{ .Values.monitoringEnabled }}
```
merlimat:
```suggestion
      annotations:
        prometheus.io/port: "{{ .Values.server.ports.metrics }}"
        prometheus.io/scrape: "{{ .Values.monitoringEnabled }}"
```

deploy/charts/oxia-controller/templates/deployment.yaml
```diff
@@ -11,6 +11,9 @@ spec:
       {{- include "oxia-controller.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      annotations:
+        prometheus.io/port: {{ .Values.ports.metrics }}
+        prometheus.io/scrape: {{ .Values.monitoringEnabled }}
```
merlimat:
```suggestion
      annotations:
        prometheus.io/port: "{{ .Values.ports.metrics }}"
        prometheus.io/scrape: "{{ .Values.monitoringEnabled }}"
```

PR #313, opened 2023-04-06
Dynamically add/remove namespaces when the cluster config is updated

merlimat:


coordinator/impl/coordinator.go
```diff
@@ -340,3 +361,67 @@ func (c *coordinator) ClusterStatus() model.ClusterStatus {
 	defer c.Unlock()
 	return *c.clusterStatus.Clone()
 }
+
+func (c *coordinator) waitForExternalEvents() {
+	refreshTimer := time.NewTicker(c.clusterConfigRefreshTime)
```
coderzc, line 366: Can we use `viper.WatchConfig()` instead of Ticker?

merlimat: That's a good point. Do you want to take this over? Perhaps we can merge this and then make the change to react instead of polling for changes.

PR #372, opened 2023-08-01
Add documentation for bare metal deployment

mattisonchao:


docs/bare-metal-deploy.md
```diff
@@ -0,0 +1,133 @@
...
+      --profile-bind-address string   Bind address for pprof (default "127.0.0.1:6060")
+```
+
+## Go for testing
+
+After all of the components are up and running without an error log. We can use oxia-perf to test. the command is as follows.
+
+```shell
+./bin/oxia perf --rate 10000
+```
```
coderzc, line 130: Do we need to specify a service address?

merlimat (approved):

PR #386, opened 2023-11-21
refactor(*): group import by standard/third/current

lsytj0413:
Group import ordering by:

```
std
<blank>
third
<blank>
project
```

There is no convention about the order of `third` and `project`, but separate from stdlib is meaningful.

This is done by blow script:

```
find . -type f -name '*.go' | grep -v '.pb.go' | xargs goimports -local=github.com/streamnative/oxia -w
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -15,3 +15,14 @@
 run:
   skip-dirs:
     - pkg/generated
+
+linters:
+  enable:
+    - goimports
+
+linters-settings:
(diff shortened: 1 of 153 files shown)


lsytj0413:
> Can this be also enforced by the linter?

Fixed, this could be done by lint auto-fix；but it will not been run automated. Another why is configuration editor's auto format on save.

merlimat (approved):

Issue #388, opened 2023-11-23
【log】replace pkg's  zerolog reference with slog


lsytj0413:
Currently we have upgrade golang's version to 1.21，we should use build in [slog](https://go.dev/blog/slog) to replace it, and make zerolog as it's handler.

For server:

- reference slog in any pkgs
- initialize slog handler with zerolog when process initialize

For client:

- only use slog
- make the user to choice handler for slog

lsytj0413:
> 👍 Looks like a good idea

Ok，i will change client first.

PR #390, opened 2023-11-28
fix(388): replace zerolog reference with slog

lsytj0413:
Fixes: https://github.com/streamnative/oxia/issues/388

### DIFF
--- a/common/client_pool.go
+++ b/common/client_pool.go
@@ -17,13 +17,12 @@ package common
 import (
 	"context"
 	"io"
+	"log/slog"
 	"sync"
 	"time"
 
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/pkg/errors"
-	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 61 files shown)


common/client_pool.go
```diff
@@ -65,10 +64,11 @@ func (cp *clientPool) Close() error {
...
 		err := cnx.(*grpc.ClientConn).Close()
 		if err != nil {
-			cp.log.Warn().
-				Str("server_address", target).
-				Err(err).
-				Msg("Failed to close GRPC connection")
+			cp.log.Warn(
+				"Failed to close GRPC connection",
+				slog.String("server_address", target),
+				slog.Any("Error", err),
```
merlimat, line 70: nit: should we consider standardizing on `error` as the key for all error messages?

merlimat (approved):
Thanks for the great refactoring!

common/client_pool.go
lsytj0413: yes，i name the key of error as `Error`，and slog didn't provide an convention for it.

merlimat: I was meaning: would it be better to use lower cap `error` to be more consistent to the other log keys?

lsytj0413: sorry, i misunderstand it. I will fix it at next time.

Issue #324, opened 2023-04-12
Failed test with leaked iterator: TestLeaderController_DeleteShard


merlimat:
```
--- FAIL: TestLeaderController_DeleteShard (0.02s)
    leader_controller_test.go:996: 
        	Error Trace:	/home/runner/work/oxia/oxia/server/leader_controller_test.go:996
        	Error:      	Received unexpected error:
        	            	leaked iterators:
        	            	(1) attached stack trace
        	            	  -- stack trace:
        	            	  | github.com/cockroachdb/pebble.(*DB).Close
        	            	  | 	/home/runner/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20221104214247-8dc60b62ebbf/db.go:1314
        	            	  | oxia/server/kv.(*Pebble).Close
        	            	  | 	/home/runner/work/oxia/oxia/server/kv/kv_pebble.go:320
        	            	  | oxia/server/kv.(*db).Close
        	            	  | 	/home/runner/work/oxia/oxia/server/kv/db.go:132
        	            	  | oxia/server.(*leaderController).close
        	            	  | 	/home/runner/work/oxia/oxia/server/leader_controller.go:830
        	            	  | oxia/server.(*leaderController).Close
        	            	  | 	/home/runner/work/oxia/oxia/server/leader_controller.go:790
        	            	  | oxia/server.TestLeaderController_DeleteShard
        	            	  | 	/home/runner/work/oxia/oxia/server/leader_controller_test.go:996
        	            	  | testing.tRunner
        	            	  | 	/opt/hostedtoolcache/go/1.19.7/x64/src/testing/testing.go:1446
        	            	  | runtime.goexit
        	            	  | 	/opt/hostedtoolcache/go/1.19.7/x64/src/runtime/asm_amd64.s:1594
        	            	Wraps: (2) leaked iterators:
        	            	  | 0.0:
        	            	  |   000004:[__oxia/term#1,SET-__oxia/term#1,SET]
        	            	Error types: (1) *withstack.withStack (2) *errutil.leafError
        	Test:       	TestLeaderController_DeleteShard
```

PR #393, opened 2023-11-30
fix(324): close channel after iterator is closed

lsytj0413:
Fixed: https://github.com/streamnative/oxia/issues/324


merlimat (approved):
Very nice find!

PR #404, opened 2023-12-09
chore(*): enable gosec/godot linter

lsytj0413:


oxia/async_client_impl.go
```diff
@@ -212,10 +212,10 @@ func (c *clientImpl) List(ctx context.Context, minKeyInclusive string, maxKeyExc
 	ch := make(chan ListResult)
 	wg := common.NewWaitGroup(len(shardIds))
 	for _, shardId := range shardIds {
-		shardIdPtr := &shardId
+		shardIdPtr := shardId
 		go func() {
 			request := &proto.ListRequest{
-				ShardId:        shardIdPtr,
+				ShardId:        &shardIdPtr,
```
merlimat, line 218: This looks like it was an actual bug, right?

merlimat (approved):

oxia/async_client_impl.go
lsytj0413: Old implement is an bug（before [go1.22 and go1.21](https://github.com/golang/go/wiki/LoopvarExperiment)）

PR #408, opened 2023-12-13
chore(*): enable more revive linter rules

lsytj0413:


lsytj0413:
Some rules have been disabled, them will enable in future pr.

PR #410, opened 2023-12-16
chore(*): enable revive.typeassert rules

lsytj0413:
Add an generic `treeMap` wrapper for `redblacktree` to simplify the wal's code to avoid use of type assert. Currently `gods` didn't have an release of generic support, after the official support of generic is released we should migrate to it.

### DIFF
--- a/server/wal/treemap.go
+++ b/server/wal/treemap.go
@@ -1,0 +1,126 @@
+// Copyright 2023 StreamNative, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


server/wal/treemap.go
```diff
@@ -0,0 +1,126 @@
...
+package wal
+
+import (
+	"fmt"
+
+	rbt "github.com/emirpasic/gods/trees/redblacktree"
+	"github.com/emirpasic/gods/utils"
+)
+
+type treeMap[K comparable, V any] struct {
```
merlimat, line 24: Would it make sense to just switch to a map implementation that already supports generics? eg: https://github.com/zyedidia/generic/tree/master/btree

lsytj0413: `gods` seems have 10x star than `zyedidia/generic`, and if we switch to `zyedidia/generic`, we should implement min、max、floor by our self.

I have searched for an generic data structure implementation in golang, and didn't found any powerful as `gods` library.

merlimat: Sounds good. Thanks for checking!

merlimat (approved):

PR #416, opened 2023-12-28
chore(*): enable revive.flag-parameter rules

lsytj0413:


cmd/client/put/cmd.go
```diff
@@ -138,17 +139,13 @@ func (Query) Unmarshal(b []byte) (common.Query, error) {
...
 
-func convertValue(binary bool, value string) ([]byte, error) {
-	if binary {
-		decoded := make([]byte, int64(float64(len(value))*0.8))
-		_, err := base64.StdEncoding.Decode(decoded, []byte(value))
-		if err != nil {
-			return nil, ErrBase64ValueInvalid
-		}
-		return decoded, nil
+func convertBinaryValue(value string) ([]byte, error) {
```
mattisonchao: convertToBinary or convertTo ? 

Because convert binary value looks like we want to convert a binary to other type. :)

lsytj0413: rename to `convertFromBinaryValue`，because when this is called, the value is an `Binary base64` format.

merlimat (approved):

PR #423, opened 2023-12-29
Add K8S startup probe

merlimat:
A fast sequence of rolling restarts can cause some temporary unavailability on some shards. 

The sequence is as this: 

 1. We start where `oxia-4` leader. `oxia-3` and `oxia-5` are followers
 2. `oxia-4` gets restarted
 3. Immediately, `oxia-5` is made leader and `oxia-3` a follower. `oxia-4` is not attached yet to the leader. The coordinator will keep trying until `oxia-4` comes back up
 4. `oxia-3` is restarted, before `oxia-4` was added as a follower 
 5. The shard is unavailable (to writes)
 6. Unailability resolves when the coordinator is getting `oxia-4` back in the followers list and the leader can make progress.

### Modifications

Add a startup probe that delays the rollout to the next pod by at least 1min. This slows down the deployment, but it greatly reduces the chances of deploying to a pod that is currently needed for the leader to make progres..

### DIFF
--- a/deploy/charts/oxia-cluster/templates/_helpers.tpl
+++ b/deploy/charts/oxia-cluster/templates/_helpers.tpl
@@ -69,3 +69,14 @@ exec:
 initialDelaySeconds: 10
 timeoutSeconds: 10
 {{- end }}
+
+{{/*
+Probe
+*/}}
+{{- define "oxia-cluster.startup-probe" -}}
+exec:
+  command: ["oxia", "health", "--port={{ . }}"]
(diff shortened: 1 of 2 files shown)


merlimat:
Yes, it's correct that this doesn't fix the problem at 100%. 

In order to fix it properly it would require to fetch the cluster status from the config map and the n probe the storage node with `GetStatus` calls to wait until the follower is attached to all the shards it needs to and also caught up with the leader. 

It's indeed quite complicated and would require doing it in an operator

Issue #407, opened 2023-12-12
【Feature】add crc to wal


lsytj0413:
oxia use wal to storage the `LogEntry` in segment files, should we add `crc` to each record to protect data corrupted？

merlimat:
Yes, one idea was to use a crc based on the past entries of the log as well. (eg: similar to git history). That would help in identifying any issues in non-common histories of the wal across nodes.

merlimat:
@lsytj0413 Perhaps we can start with a proposal and design doc for how to implement the change. We can use discussions. eg: https://github.com/streamnative/oxia/discussions/new?category=proposal

lsytj0413:
> @lsytj0413 Perhaps we can start with a proposal and design doc for how to implement the change. We can use discussions. eg: https://github.com/streamnative/oxia/discussions/new?category=proposal

Yes, i will open an proposal to describe the detail design first.

lsytj0413:
> @lsytj0413 Perhaps we can start with a proposal and design doc for how to implement the change. We can use discussions. eg: https://github.com/streamnative/oxia/discussions/new?category=proposal

Propocal at https://github.com/streamnative/oxia/discussions/412，PTAL

lsytj0413:
Seems this have been implemented at https://github.com/streamnative/oxia/pull/538#event-14438514703
@merlimat  this can be close？

PR #425, opened 2024-01-02
WIP: fix(407): add crc to wal entry

lsytj0413:
Fixed：https://github.com/streamnative/oxia/issues/407

First implement of add `crc` to `wal` entry, currently there is a lot of work should be done:

- Implement `crc` reset in TruncateLog
- validate `crc` value when read data
- add `decoder` to simplify code
- add examples

and so on.


lsytj0413:
Close because https://github.com/streamnative/oxia/pull/538 already implement it.

PR #363, opened 2023-07-11
Fixed deadlock in the leader controller wal batcher

merlimat:
The batcher component that aggregates multiple write requests in the leader controller, before passing them to the WAL, has the potential for deadlocks.

Example:

This go-routine has acquired the leader controller lock and it's now block on the batcher channel (default size is 1, though increasing the channel size won't fix the core problem).

```
1 @ 0x4482d6 0x4116ec 0x41129d 0x9dc4f2 0x1980eca 0x1980ae5 0x198093e 0x1982d45 0x991678 0x985707 0x991538 0x96e8b0 0x97396f 0x96c198 0x479ae1
# labels: {"bind":"[::]:6648", "oxia":"public"}
#	0x9dc4f1	oxia/common/batch.(*batcherImpl).Add+0x51							/src/oxia/common/batch/batcher.go:51
#	0x1980ec9	oxia/server.(*leaderController).appendToWal+0x169						/src/oxia/server/leader_controller.go:688
#	0x1980ae4	oxia/server.(*leaderController).write+0x124							/src/oxia/server/leader_controller.go:668
#	0x198093d	oxia/server.(*leaderController).Write+0x7d							/src/oxia/server/leader_controller.go:655
#	0x1982d44	oxia/server.(*publicRpcServer).Write+0xe4							/src/oxia/server/public_rpc_server.go:89
#	0x991677	oxia/proto._OxiaClient_Write_Handler.func1+0x77							/src/oxia/proto/client_grpc.pb.go:351
#	0x985706	github.com/grpc-ecosystem/go-grpc-prometheus.(*ServerMetrics).UnaryServerInterceptor.func1+0x86	/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-prometheus@v1.2.0/server_metrics.go:107
#	0x991537	oxia/proto._OxiaClient_Write_Handler+0x137							/src/oxia/proto/client_grpc.pb.go:353
#	0x96e8af	google.golang.org/grpc.(*Server).processUnaryRPC+0xdef						/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:1345
#	0x97396e	google.golang.org/grpc.(*Server).handleStream+0xa2e						/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:1722
#	0x96c197	google.golang.org/grpc.(*Server).serveStreams.func1.2+0x97					/go/pkg/mod/google.golang.org/grpc@v1.54.0/server.go:966
```

[...]

PR #431, opened 2024-01-17
Fixed create and modified timestamps not getting set

merlimat:
After #363, the timestamps were always getting set to 0


RobertIndie (approved):
We also need to modify the expected result here: https://github.com/streamnative/oxia/blob/0cd9f4fb18688ef19cf14674a5f5b1311f99d0ed/oxia/example_test.go#L81

Issue #455, opened 2024-04-28
Reserve more metadata in coordinator status configmap


labuladong:
https://github.com/streamnative/oxia/blob/281313c5be96734e50b3e17e89ed4e5179b0d9bf/coordinator/impl/metadata_configmap.go#L133-L138

This part of code will always create a new configmap and override the old one. But I think it should reserve the metadata part of the old configmap, for example ownerReferences filed is useful for cascading deletion. The labels field may also be useful.

PR #456, opened 2024-04-28
improve oxia coordinatior status configmap

labuladong:
Fixes https://github.com/streamnative/oxia/issues/455


mattisonchao:
Hi @labuladong  
do we have any way to let is support server-side patches?

coordinator/impl/k8s_client.go
```diff
@@ -85,16 +91,23 @@ type clientImpl[Resource resource] struct {
 	clientFunc func(string) ResourceInterface[Resource]
 }
 
-func (c *clientImpl[Resource]) Upsert(namespace string, resource *Resource) (result *Resource, err error) {
+func (c *clientImpl[Resource]) Upsert(namespace, name string, resource *Resource) (*Resource, error) {
```
mattisonchao, line 94: Sorry, I am considering this name. upsert means update and insert. :/ is it okay to use a patch inside this function?

labuladong: I think it's ok. patch is one kind of update, right?

mattisonchao (approved):

PR #492, opened 2024-06-07
tls: support server TLS options

mattisonchao:
## Motivation
Follow up the last TLS support PR. This PR is trying to support CLI to set TLS options for Server

## Modification

- Add new command parameters to set TLS options.

## Others
- We might consider using different TLS options for public/internal; we can easily support it in the future if we need it.

### DIFF
--- a/cmd/server/cmd.go
+++ b/cmd/server/cmd.go
@@ -15,9 +15,10 @@
 package server
 
 import (
-	"github.com/streamnative/oxia/common/security"
 	"io"
 	"time"
+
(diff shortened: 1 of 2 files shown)


cmd/server/cmd.go
```diff
@@ -18,6 +18,8 @@ import (
 	"io"
 	"time"
 
+	"github.com/streamnative/oxia/common/security"
```
lsytj0413: Run `make lint` will group `github.com/streamnative/oxia`'s import to same order.

merlimat (approved):

PR #495, opened 2024-06-13
coordinator: fix too much cursor causes by ensemble breaking

mattisonchao:
### Motivation

failed test: https://github.com/streamnative/oxia/actions/runs/9489462225/job/26150775945

In the old logic, we count the removed node into the ensemble, which causes some issues.

1. the `fencingQuorum` is breaking the expected quorum number.
2. assign more followers to the shared leader(since we are waiting `quorumFencingGracePeriod`). Which will cause `too many cursor` exceptions and let the leader election constantly fail. 

### Alternative

I am unsure why we must add `RemovedNodes` to the quorum. If we must have a reason, we can filter it in `quorumFencingGracePeriod` logic. :)

### DIFF
--- a/coordinator/impl/shard_controller.go
+++ b/coordinator/impl/shard_controller.go
@@ -502,8 +502,8 @@ func (s *shardController) newTermQuorum() (map[model.ServerAddress]*proto.EntryI
 	timer := s.newTermQuorumLatency.Timer()
 
 	ensembleQuorum := s.shardMetadata.Ensemble
-	fencingQuorumSize := len(ensembleQuorum)
-	majority := fencingQuorumSize/2 + 1
+	ensembleQuorumSize := len(ensembleQuorum)
+	majority := ensembleQuorumSize/2 + 1
 
 	// Use a new context, so we can cancel the pending requests
 	ctx, cancel := context.WithCancel(s.ctx)
@@ -514,7 +514,7 @@ func (s *shardController) newTermQuorum() (map[model.ServerAddress]*proto.EntryI
 		model.ServerAddress
 		*proto.EntryId
 		error
-	}, fencingQuorumSize)
+	}, ensembleQuorumSize)
 
 	for _, sa := range ensembleQuorum {
 		// We need to save the address because it gets modified in the loop
@@ -557,7 +557,7 @@ func (s *shardController) newTermQuorum() (map[model.ServerAddress]*proto.EntryI
 	var err error
 
 	// Wait for a majority to respond
-	for successResponses < majority && totalResponses < fencingQuorumSize {
+	for successResponses < majority && totalResponses < ensembleQuorumSize {
 		r := <-ch
 
 		totalResponses++
@@ -576,7 +576,7 @@ func (s *shardController) newTermQuorum() (map[model.ServerAddress]*proto.EntryI
 
 	// If we have already reached a quorum of successful responses, we can wait a
 	// tiny bit more, to allow time for all the "healthy" nodes to respond.
-	for err == nil && totalResponses < fencingQuorumSize {
+	for err == nil && totalResponses < ensembleQuorumSize {
 		select {
 		case r := <-ch:
 			totalResponses++


coordinator/impl/shard_controller.go
```diff
@@ -368,6 +368,10 @@ func (s *shardController) electLeader() error {
 
 func (s *shardController) deletingRemovedNodes() error {
 	for _, ds := range s.shardMetadata.RemovedNodes {
+		// todo: merge to single request
+		if _, err := s.newTerm(s.ctx, ds); err != nil {
```
merlimat, line 372: Why do we need to trigger a new term when removing nodes?

merlimat, line 372: Why do we need to trigger a new term when removing nodes?

coordinator/impl/shard_controller.go
```diff
@@ -497,9 +501,9 @@ func (s *shardController) internalNewTermAndAddFollower(ctx context.Context, nod
 func (s *shardController) newTermQuorum() (map[model.ServerAddress]*proto.EntryId, error) { //nolint:revive
 	timer := s.newTermQuorumLatency.Timer()
 
-	fencingQuorum := mergeLists(s.shardMetadata.Ensemble, s.shardMetadata.RemovedNodes)
-	fencingQuorumSize := len(fencingQuorum)
-	majority := fencingQuorumSize/2 + 1
+	ensembleQuorum := s.shardMetadata.Ensemble
```
merlimat, line 504: The fencing quorum here should also include the removed nodes. We need to make sure there is not another coordinator instance that will fence them in a different term with a different majority.

merlimat (commented):
Is the effect that we have many repeated entries in the cluster status? I think I've seen that problem at some point though I was unable to replicate again.

coordinator/impl/shard_controller.go
mattisonchao: We have a check on the server side. The follower will reject the deleting request if the term does not fit.

mattisonchao:
>Is the effect that we have many repeated entries in the cluster status? I think I've seen that problem at some point though I was unable to replicate again.

Could you please clarify the repeated entries in cluster status? I am unsure which part you mentioned. 

I tested some times. I found our leader election is not stable after introducing the node swap function, which will cause the concurrent leader election. you can simply close server 1 after you change configuration here. 
https://github.com/streamnative/oxia/blob/9696873987158b171c9ffe5d825571c00fff271f/coordinator/impl/coordinator_e2e_test.go#L570

some issues might happen when you run it many times. 

The RC should be related to concurrent leader election in the shards controller.  

1. even though we cancelled the old context, we didn't interrupt the old goroutine. so the behaviour will be undefined. plus, our context is not accurate, and it can make things worse. 

e.g: https://github.com/streamnative/oxia/blob/9696873987158b171c9ffe5d825571c00fff271f/coordinator/impl/shard_controller.go#L289 
If we are using this context to control the concurrent, we might need to assign this to all the sub-operations in the election. but I suggest to use a `channel` to sequence operation and make it simple.

2. if the removed node is crashed. the leader election will never done.  (caused by delete shards error)
https://github.com/streamnative/oxia/blob/9696873987158b171c9ffe5d825571c00fff271f/coordinator/impl/shard_controller.go#L341

etc... 

Sorry,  I just brought up some issues I found but still have no time to fix them. also, it's not very common and urgent. I guess we can fix it later.

coordinator/impl/shard_controller.go
mattisonchao: >The fencing quorum here should also include the removed nodes. We need to make sure there is not another coordinator instance that will fence them in a different term with a different majority.

1. Do you mean we might have another coordinator deployment?  
2. Yes, we can keep the logic here and just filter all the unexpected followers. but I am considering if it's breaking the expected quorum. 
for example: 
we have server [sv1,sv2,sv3] and deleted node [sv4].  the swap operation is from `sv4` to `sv3`.  and right now. The `sv3` and `sv4` crashed at the same time.  
then the quorum should be `4/2 +1 = 3` success response. but two crashed servers will let the election constantly fail. 
However, if we keep the quorum equal to `3/2 + 1 =2` success response (don't count removed node). even `sv4` and `sv3` crashed at same time. the leader election can still succeed.

PR #496, opened 2024-06-24
auth: support OIDC authentication provider and token authentication

mattisonchao:
### Motivation

We should support some general authentication methods for the Oxia server for security concerns.

### Modification

- Support basic OIDC authentication provider on the server
- Support basic token authentication on the client
- Support integration test for OIDC with the token.

### Next

- Performance will be affected Since we use GRPC per call authentication. We might do a cache for the token to improve the performance.
- For next improvement. We can customise a transport layer authentication. But we already have mTLS. We can evaluate it. :)


server/auth/oidc.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// any of the client audience in the allowed is passed
+	audienceAllowed := false
+	audienceArr := idToken.Audience
+	for _, audience := range audienceArr {
+		if _, ok := p.allowedAudiences[audience]; ok {
+			audienceAllowed = true
+		}
+	}
```
tuteng, line 128: Do you think it's necessary to check the `permissions/scope` field to verify the permissions?

tuteng (approved):

server/auth/interceptor.go
```diff
@@ -0,0 +1,105 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	MetadataAuthorizationKey = "authorization"
+	TokenPrefix              = "Bearer "
+)
```
tuteng, line 33: Maybe we should accept the header of authorizaiton/Authorizaiton and Bearer/bearer

tuteng (approved):

server/auth/oidc.go
mattisonchao: we don't support authorization yet. We can consider it when we support authorization. :)

server/auth/interceptor.go
mattisonchao: Okay, make sense. let me add it later. :)

mattisonchao: Sorry, the `authorizaiton` and `Bearer` should be standard. We can support leveraging case insensitivity in other PRs

merlimat (approved):
Nice work!  👍

PR #500, opened 2024-07-31
server: enable pebble bloom filter for better `read` performance

mattisonchao:
## Motivation

Enable table bloom filter to improve the `read` performance of oxia. The reason is that the bloom filter can reduce the disk IO and quickly judge if the current block has this key. 

## Benchmark Comparing

Env:  
- 2 core 4G, 2G block cache. 100G gp3 SSD
- 64 Bytes 3 Shards
- left: bloom enabled, right: bloom disabled

### Insert only (sequential key distribution)

image

### YCSB A (50% read, 50% write, Zipfian key distribution)

image

### YCSB B (%5 write, 95% read, Zipfian key distribution)

image

### YCSB C (100% read, Zipfian key distribution)

image

### DIFF
--- a/server/kv/kv_pebble.go
+++ b/server/kv/kv_pebble.go
@@ -16,6 +16,7 @@ package kv
 
 import (
 	"fmt"
+	"github.com/cockroachdb/pebble/bloom"
 	"io"
 	"log/slog"
 	"os"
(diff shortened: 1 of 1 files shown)


merlimat:
> left: bloom enabled, right: bloom disabled

@mattisonchao Are the charts before or after the change? left vs right looks like two different charts, throughput & latency

merlimat (approved):

PR #511, opened 2024-09-12
Rename ShardId field to Shard

merlimat:
Renaming the protobuf field from `ShardId` to `Shard`. 

The motivation is to have consistent `shard: 123` in the logs, such that filtering them based on shard becomes easier.

All logs context are already using the `shard` label, though when we print Protobuf struct directly, it would today have `shardId`.


mattisonchao:
We should mention that is breaking changes.

merlimat:
> We should mention that is breaking changes.

No, this is not changing the wire protocol. Just the name of the structures in Go. The fields are identified by their tag number in Protobuf, not the field name.

mattisonchao (approved):

PR #522, opened 2024-09-20
Add pprof flag for the server and coordinator

zymap:
---

### Motivation

Enable the pprof to debug issues.


mattisonchao:
This command already supported.

PR #530, opened 2024-09-24
feat(cmd): support wal cmd `truncate` command

mattisonchao:
### Motivation

Introduce the wal cmd `truncate` command to truncate corrupted unsynced data in the write-ahead log before implementing a CRC-based write-ahead log.

### DIFF
--- a/cmd/wal/truncate/cmd.go
+++ b/cmd/wal/truncate/cmd.go
@@ -16,16 +16,13 @@ package truncate
 
 import (
 	"github.com/spf13/cobra"
+	common "github.com/streamnative/oxia/cmd/wal/common"
 	"github.com/streamnative/oxia/server/wal"
 	"log/slog"
 	"math"
 )
 
 type truncateOptions struct {
-	namespace string
-	shard     int64
-	walDir    string
-
 	lastEntry      bool
 	safePointEntry int64
 }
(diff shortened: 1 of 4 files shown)


cmd/wal/truncate/cmd.go
```diff
@@ -0,0 +1,103 @@
...
+		Short: "truncate the WAL",
+		Long:  `truncate the WAL by some conditions`,
+		RunE:  exec,
+	}
+)
+
+func init() {
+	Cmd.Flags().Int64Var(&options.shard, "shard", 0, "shard id")
+	Cmd.Flags().StringVar(&options.namespace, "namespace", "default", "namespace name")
+	Cmd.Flags().StringVar(&options.walDir, "wal-dir", "", "directory path")
```
merlimat: These should be options of the `wal` command, so they get inherited by all subcommands

cmd/wal/truncate/cmd.go
```diff
@@ -0,0 +1,103 @@
...
+	// operations
+	Cmd.Flags().Int64Var(&options.safePointEntry, "safe-point-entry", math.MaxInt64, "the last safe entry offset")
+	Cmd.Flags().BoolVar(&options.lastEntry, "last-entry", false, "if trim the last entry")
```
merlimat: Maybe rename into
```suggestion
	Cmd.Flags().BoolVar(&options.lastEntry, "trim-last-entry", false, "trim the last entry from the wal")
```

cmd/wal/truncate/cmd.go
mattisonchao: applied.

cmd/wal/truncate/cmd.go
mattisonchao: Refined the usage and changed the prefix to `truncate`.

merlimat (approved):

PR #532, opened 2024-09-24
On delete-range send 1 single notification to clients

merlimat:
When there is a delete-range request, we are currently sending 1 `deleted` notification for each of the keys that were deleted in the range. 

The problem is that the number of keys in the range is unbounded and can lead to a huge batch in the DB and also to a big notification batch to send out to client. It's not easy to split the notification batch, since it needs to be written atomically with the db update itself.

Instead, we add a new notification type that will simply tell the client that a range of keys was deleted.


server/kv/db.go
```diff
@@ -595,27 +596,26 @@ func (d *db) applyDelete(batch WriteBatch, notifications *notifications, delReq
...
+		notifications.DeletedRange(delReq.StartInclusive, delReq.EndExclusive)
 	}
 
-	for it.Next() {
-		if notifications != nil {
-			notifications.Deleted(it.Key())
-		}
-		err := updateOperationCallback.OnDelete(batch, it.Key())
+	if updateOperationCallback != NoOpCallback {
+		it, err := batch.KeyRangeScan(delReq.StartInclusive, delReq.EndExclusive)
```
mattisonchao, line 604: someone might add a new key in this range?

mattisonchao (approved):

PR #538, opened 2024-09-26
feat(wal): support chained CRC for write ahead log

mattisonchao:
### Motivation

To prevent data corruption from affecting the stability or causing unrecoverable exceptions. Introduce CRC to help validate if the data has corrupted.

### Modification

- Support chained CRC `CRC(n) = CRC( DATAn, CRCn-1 )` we can validate that followers have exactly the same identical WAL.
- Support `TxnFormatVersion2` for compatibility with the existing format.
- Adding CRC WAL test

### Others

will add the proposal later.

### DIFF
--- a/server/util/crc/crc.go
+++ b/server/util/crc/crc.go
@@ -18,6 +18,8 @@ import (
 	"hash/crc32"
 )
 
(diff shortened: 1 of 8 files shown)


merlimat (approved):
👍

PR #548, opened 2024-10-09
feat(cli): support list ephemeral keys and pass multi keys to delete

mattisonchao:
### Motivation

Support list ephemeral keys and pass multi keys to delete operation. Which can be used for uncleaned ephemeral keys by bugs.

example of deletion:
```
./oxia  client list --ephemeral -n xxxx | xargs ./oxia client delete -n xxxx
```

example of get shadow keys

```
oxia client list --ephemeral -n xxxx | xargs ./oxia client get --ephemeral-shadow-key -n xxxx
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -92,7 +92,7 @@ linters-settings:
         severity: warning
         disabled: false
         arguments:
-          - 15
+          - 17
       - name: cyclomatic
         severity: warning
         disabled: false
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 .PHONY: build
 build:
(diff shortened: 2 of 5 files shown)


merlimat:
I would defer this feature to have a more general approach based on secondary indexes, and not expose the `shadow-key` in the tools.

Session manager can now be implemented on top of secondary indexes, so that we can have a single implementation. We just need to ensure proper migration of existing session during an upgrade.

It's already possible to get a list of the internal sessions states right now (by using `--end-key __oxia/session/999999999` kind of query (even though it's not very nice).

mattisonchao:
@merlimat 
Yep, we can redesign the CLI by implementing the secondary index-based session manager.

PR #551, opened 2024-10-15
fix(server): support commit context to help auto-correct internal generated value

mattisonchao:
## Motivation

#505  introduced breaking changes when upgrading your cluster from the old to the new version. If you do the rolling upgrade, the internal version generator will use the wrong initial value for entries.

The main idea here is that we can use the minimal changes to help auto-correct the generated value between shards due to only upgrading affection.

## Modification

- Introduce a new concept that named `commitContext`. We will pass this context when we create a replication stream from leader to follower. Record the commit offset and the last version in that time. Therefore, the follower can understand the correct version id at the specific commitOffset and auto-override the invalid value to reach the consensus.


mattisonchao:
We need to find a way to be compatible with new features and introduce a tool for this kind of recovery.

PR #595, opened 2025-01-20
test(572): fix flaky test with TestSecondaryIndices_List

lsytj0413:
Before fix，when run with `go test -timeout 10m -run ^TestSecondaryIndices_List$ github.com/streamnative/oxia/server -gcflags all='-N' -v -count=10000`, the flaky test reproduced.

After fix, run multi times tests without fail.

fix: https://github.com/streamnative/oxia/issues/582


lsytj0413:
We should work on a mechanism to avoid close database when any iterator or anything is working on it.

mattisonchao (approved):

Issue #596, opened 2025-01-23
bug: panic: runtime error: invalid memory address or nil pointer dereference


mattisonchao:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x192d87d]

goroutine 104 [running]:
github.com/streamnative/oxia/coordinator/impl.(*shardController).internalNewTermAndAddFollower(0xc0001d0b00, {0x2534a60?, 0xc000af2730?}, {{0xc00011dea0, 0x45}, {0xc000516cc0, 0x2f}}, 0xc0005d73b0)
	/src/oxia/coordinator/impl/shard_controller.go:513 +0xfd
github.com/streamnative/oxia/coordinator/impl.(*shardController).run(0xc0001d0b00)
	/src/oxia/coordinator/impl/shard_controller.go:202 +0x545
github.com/streamnative/oxia/common.DoWithLabels.func1({0x2534a28?, 0xc00077f890?})
	/src/oxia/common/pprof.go:46 +0x13
runtime/pprof.Do({0x2534a60?, 0xc000139950?}, {{0xc00068eb40?, 0x0?, 0x0?}}, 0xc000484f30)
	/usr/local/go/src/runtime/pprof/runtime.go:51 +0x8c
github.com/streamnative/oxia/common.DoWithLabels({0x2534a60, 0xc000139950}, 0xc000644210, 0xc00063e1d0)
	/src/oxia/common/pprof.go:42 +0x34d
created by github.com/streamnative/oxia/coordinator/impl.NewShardController in goroutine 1
	/src/oxia/coordinator/impl/shard_controller.go:150 +0xba5
```

PR #599, opened 2025-01-24
test(596): fix panic when duplicate fence

lsytj0413:
Fix：https://github.com/streamnative/oxia/issues/596

it looks like the code and comment didn't match, and we should fix the code.

### DIFF
--- a/server/follower_controller.go
+++ b/server/follower_controller.go
@@ -276,7 +276,7 @@ func (fc *followerController) NewTerm(req *proto.NewTermRequest) (*proto.NewTerm
 			slog.Int64("new-term", req.Term),
 			slog.Any("status", fc.status),
 		)
-		return nil, common.ErrorFollowerAlreadyFenced
+		return nil, common.ErrorInvalidStatus
 	}
 
 	if fc.db == nil {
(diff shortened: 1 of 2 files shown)


mattisonchao:
Hi, @lsytj0413 
Sorry, I don't have time to review it right now. but if you are right we might need to revert #573.

cc @merlimat

merlimat:
I don't think we should revert #573 as it was solving a specific issue :). The coordinator needs to know when to stop fencing a node that was already successfully fenced (and maybe the ack was lost or timed-out). 

@lsytj0413  Can you clarify how this affected #596 and causes the panic?

mattisonchao:
Just a quick review of the issue.

The panic should be thrown by nil `fr`. 

https://github.com/streamnative/oxia/blob/fa6caa1cf66d0b8a2cfbd7c5e7b119e0217793f7/coordinator/impl/shard_controller.go#L513

merlimat:
Got it. So the problem was the test was not complete in the coordinator path. When the `AlreadyFenced` happens, the term is not returned, though the coordinator should actually know the last entry

merlimat:
Pushed a fix in #602

lsytj0413:
> Got it. So the problem was the test was not complete in the coordinator path. When the `AlreadyFenced` happens, the term is not returned, though the coordinator should actually know the last entry

yes, when the follower is AlreadyFenced，the term and last entry is not returned；we can make NewTerm idempotent, or just fix the ErrCode.

PR #606, opened 2025-02-17
feat: support leader election print leader entry id

mattisonchao:


.github/workflows/docker_publish.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
-          platforms: linux/x86_64,linux/arm64
+          platforms: linux/x86_64
```
merlimat, line 61: This is needed to publish multi-arch image

mattisonchao: Yes, I just use this branch to build a testing image. :)  btw, I should close this PR. sorry.

PR #608, opened 2025-02-19
fix(server): fail pending request when `quorumAckTracker` closed

mattisonchao:
### Motivation

This is the follow-up PR for #607. We should fail the pending request when `quorumAckTracker` closes.

### Modification

- Introduce a callback
- fail pending request when `quorumAckTracker` closes.

### DIFF
--- a/common/callback/once.go
+++ b/common/callback/once.go
@@ -1,3 +1,17 @@
+// Copyright 2025 StreamNative, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


common/callback/once.go
```diff
@@ -0,0 +1,63 @@
...
+//
+// The generic type T represents the result type of the operation, and the callbacks
+// provide the behavior for handling success or failure of the operation.
+//
+// Fields:
+// - OnComplete: A function that gets called with the result of type T when the operation completes successfully.
+// - OnCompleteError: A function that gets called with an error if the operation fails.
+// - completed: An atomic boolean used to track if the operation has already completed, ensuring only one callback is executed.
+
+type Once[T any] struct {
```
merlimat, line 31: Even if it's very simple, it might be good to have a simple unit test for this

merlimat (approved):

common/callback/once.go
mattisonchao: sure, let me add one.

mattisonchao: added.

Issue #594, opened 2025-01-17
bug: Failed to trim the wal


mattisonchao:
### Motivation

```
{"level":"error","time":"2025-01-17T02:20:21.045523933Z","component":"wal-trimmer","error":{"error":"failed to read from wal at offset 14835125 first=14835125 last=14877701: failed to open segment txn file /data/wal/test/shard-13/14835125.txnx: open /data/wal/test/shard-13/14835125.txnx: no such file or directory","kind":"*errors.withStack","stack":null},"namespace":"test","shard":13,"time":"2025-01-17T02:20:21.045557449Z","message":"Failed to trim the wal"}
{"level":"error","time":"2025-01-17T02:20:21.380449854Z","component":"wal-trimmer","error":{"error":"failed to read from wal at offset 13478161 first=13478161 last=13549482: failed to open segment txn file /data/wal/test/shard-15/13478161.txnx: open /data/wal/test/shard-15/13478161.txnx: no such file or directory","kind":"*errors.withStack","stack":null},"namespace":"test","shard":15,"time":"2025-01-17T02:20:21.380492941Z","message":"Failed to trim the wal"}
{"level":"error","time":"2025-01-17T02:20:21.391136207Z","component":"wal-trimmer","error":{"error":"failed to read from wal at offset 15280799 first=15280799 last=15417375: failed to open segment txn file /data/wal/test/shard-23/15280799.txnx: open /data/wal/test/shard-23/15280799.txnx: no such file or directory","kind":"*errors.withStack","stack":null},"namespace":"test","shard":23,"time":"2025-01-17T02:20:21.391163203Z","message":"Failed to trim the wal"}
{"level":"error","time":"2025-01-17T02:20:21.511553911Z","component":"wal-trimmer","error":{"error":"failed to read from wal at offset 14969597 first=14969597 last=15074458: failed to open segment txn file /data/wal/test/shard-32/14969597.txnx: open /data/wal/test/shard-32/14969597.txnx: no such file or directory","kind":"*errors.withStack","stack":null},"namespace":"test","shard":32,"time":"2025-01-17T02:20:21.51160027Z","message":"Failed to trim the wal"}
[...]

PR #617, opened 2025-02-25
fix(594): ignore txnFile not exists when trim segments

lsytj0413:
Fix: https://github.com/streamnative/oxia/issues/594


mattisonchao:
Let us consider improve this part to delete the file directly without recovery. 

cc @merlimat

lsytj0413:
> Let us consider improving this part to delete the file directly without recovery. (that would be more simpler)
> 
> cc @merlimat

Yes, delete the file is more simple and directly（i leave it on `TODO`）; at this PR fix the issue the primary goal.

mattisonchao (approved):

Issue #620, opened 2025-02-25
flaky-test: TestCoordinator_DeleteNamespace panic: metadata bad version [recovered]


lsytj0413:
```
--- FAIL: TestCoordinator_DeleteNamespace (1.07s)
panic: metadata bad version [recovered]
	panic: metadata bad version

goroutine 1546 [running]:
testing.tRunner.func1.2({0x2ffcb60, 0xc000528318})
	/opt/hostedtoolcache/go/1.23.6/x64/src/testing/testing.go:1632 +0x3fc
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.23.6/x64/src/testing/testing.go:1635 +0x6b6
panic({0x2ffcb60?, 0xc000528318?})
	/opt/hostedtoolcache/go/1.23.6/x64/src/runtime/panic.go:785 +0x132
github.com/streamnative/oxia/coordinator/impl.(*metadataProviderMemory).Store(0xc0002ac340, 0xc0009f0678, {0x32f2424, 0x1})
	/home/runner/work/oxia/oxia/coordinator/impl/metadata_memory.go:55 +0x39d
github.com/streamnative/oxia/coordinator/impl.(*coordinator).applyNewClusterConfig(0xc0004e94a0)
	/home/runner/work/oxia/oxia/coordinator/impl/coordinator.go:231 +0x370
github.com/streamnative/oxia/coordinator/impl.NewCoordinator({0x3694148, 0xc0002ac340}, 0xc000d21240, 0x0, {0x36ad090, 0xc000d89510})
	/home/runner/work/oxia/oxia/coordinator/impl/coordinator.go:139 +0xccb
github.com/streamnative/oxia/coordinator/impl.TestCoordinator_DeleteNamespace(0xc000016820)
	/home/runner/work/oxia/oxia/coordinator/impl/coordinator_e2e_test.go:405 +0x105e
testing.tRunner(0xc000016820, 0x33de368)
	/opt/hostedtoolcache/go/1.23.6/x64/src/testing/testing.go:1690 +0x227
created by testing.(*T).Run in goroutine 1
	/opt/hostedtoolcache/go/1.23.6/x64/src/testing/testing.go:1743 +0x826
FAIL	github.com/streamnative/oxia/coordinator/impl	5.747s
```

logs: https://github.com/streamnative/oxia/actions/runs/13512322219/job/37754747454?pr=617

PR #622, opened 2025-02-25
fix(619): wait shard controller loop exit when close it

lsytj0413:
Fix: https://github.com/streamnative/oxia/issues/620

When shard controller Close returned, the `run goroutine` maybe still running, it will update metadata.


lsytj0413:
We have more todo:

- when metadata version conflict, we should not panic
- when all controller Close returned, it's sub goroutine must exit
- when coordinator is closed, it should accept any function invoking like `ElectedLeader`

mattisonchao:
It seems like the issue only happened in the test because we won't run two coordinators in the same process. But it's still good for robustness. 

also, we have several background goroutines that need the same approaches. Maybe we can create a method or struct like Java `Thread` or Rust `JoinHandle`? We can have the ability to join that goroutine then.  :)

lsytj0413:
> It seems like the issue only happened in the test because we won't run two coordinators in the same process. But it's still good for robustness.
> 
> also, we have several background goroutines that need the same approaches. Maybe we can create a method or struct like Java `Thread` or Rust `JoinHandle`? We can have the ability to join that goroutine then. (maybe it already has a similar lib in the open-source world? ) :)

Yep, at next i will introduce and `JoinHandle` style to ensure all background goroutine is exited when controller close, i didn't find any open-source lib to resolve this problem, if you find anyone, i'd like to know it.

lsytj0413:
/close

use another branch.

Issue #619, opened 2025-02-25
Bug: Panic when append entry to ReadWriteSegment


lsytj0413:
Example code:

```
func TestSegmentAppendPanic(t *testing.T) {
	basePath := t.TempDir()
	rw, err := newReadWriteSegment(basePath, 0, 1024, 0, nil)
	assert.NoError(t, err)
	for i := int64(0); i < 100; i++ {
		err := rw.Append(i, fmt.Appendf(nil, "entry-%d", i))
		assert.NoError(t, err)
	}
}
```

will panic with:

```
panic: runtime error: slice bounds out of range [1030:1024] [recovered]
	panic: runtime error: slice bounds out of range [1030:1024]

goroutine 8 [running]:
testing.tRunner.func1.2({0x105424720, 0x14000034930})
	/Users/bytedance/.gvm/gos/go1.23.5/src/testing/testing.go:1632 +0x3c8
testing.tRunner.func1()
	/Users/bytedance/.gvm/gos/go1.23.5/src/testing/testing.go:1635 +0x6f8
panic({0x105424720?, 0x14000034930?})
	/Users/bytedance/.gvm/gos/go1.23.5/src/runtime/panic.go:785 +0x124
github.com/streamnative/oxia/server/wal/codec.(*V2).WriteRecord(0x1059de920, {0x105dfc000, 0x400, 0x400}, 0x406, 0x4c6ae8db, {0x14000292560, 0x8, 0x8})
	/Users/bytedance/workspace/github.com/streamnative/oxia/server/wal/codec/v2.go:145 +0x548
github.com/streamnative/oxia/server/wal.(*readWriteSegment).Append(0x140001ec160, 0x34, {0x14000292560, 0x8, 0x8})
	/Users/bytedance/workspace/github.com/streamnative/oxia/server/wal/wal_rw_segment.go:163 +0x1c8
```

We should not panic even if the segment cann't accept entry exceeded it's size.

PR #627, opened 2025-02-26
fix(619): append into ReadWriteSegment when it exceeded size should not panic

lsytj0413:
Fix: https://github.com/streamnative/oxia/issues/619

### DIFF
--- a/server/wal/wal_rw_segment.go
+++ b/server/wal/wal_rw_segment.go
@@ -154,6 +154,9 @@ func (ms *readWriteSegment) Append(offset int64, data []byte) error {
 	if len(data) == 0 {
 		return codec.ErrEmptyPayload
 	}
+	if !ms.HasSpace(len(data)) {
+		return ErrSegmentFull
+	}
 	if offset != ms.lastOffset+1 {
 		return ErrInvalidNextOffset
 	}
--- a/server/wal/wal_rw_segment_test.go
+++ b/server/wal/wal_rw_segment_test.go
@@ -16,6 +16,7 @@ package wal
 
 import (
 	"encoding/binary"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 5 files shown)


server/wal/wal_rw_segment.go
```diff
@@ -154,6 +154,9 @@ func (ms *readWriteSegment) Append(offset int64, data []byte) error {
 	if len(data) == 0 {
 		return codec.ErrEmptyPayload
 	}
+	if !ms.HasSpace(len(data)) {
```
mattisonchao, line 157: It's an extra cost for validation, and we should avoid using it without `HasSpace` check. :) 

On the other hand, the panic is enough for correctness.

mattisonchao: cc @merlimat

mattisonchao: Or we can make the `HasSpace` check inside the append and rely on the error to roll over the segment in the WAL.

merlimat: The `HasSpace()` is already checked with the WAL lock taken. I don't know how this condition could happen here. 

https://github.com/streamnative/oxia/blob/86cbb6a182c16198a69b7ec408aa04855eea27e7/server/wal/wal_impl.go#L287

lsytj0413: > The `HasSpace()` is already checked with the WAL lock taken. I don't know how this condition could happen here.
> 
> https://github.com/streamnative/oxia/blob/86cbb6a182c16198a69b7ec408aa04855eea27e7/server/wal/wal_impl.go#L287

1. This check should be in Segment append function, Because this object is the one that holds the length, and it can also prevent the situation where no inspection is carried out due to the failure of the caller.
2. panic maybe cause entire program exit, is this an elegant way to exit？

merlimat: > This check should be in Segment append function, Because this object is the one that holds the length, and it can also prevent the situation where no inspection is carried out due to the failure of the caller.

Yes, we could change the logic and return `errSegmentFull` on the append and avoid checking `HasSpace()` from outside.

> panic maybe cause entire program exit, is this an elegant way to exit？

No, though `panic()` is a good measure to protect against unexpected conditions. 

In this case I'm still not sure how the condition happens. If we check the segment size before, and we hold a lock on the whole WAL, what is the sequence of events that right now can lead to panic?

lsytj0413: Currently their is no sequence will cause append to segment exceeded size using WAL interface only, i just found it when writing unittest directly using segment struct.

I have move the logic only in segment.append function, PTAL

server/wal/wal_rw_segment_test.go
```diff
@@ -296,3 +297,16 @@ func TestReadWriteSegment_BrokenCommittedData_ErrDataCorrupted(t *testing.T) {
...
+	basePath := t.TempDir()
+	rw, err := newReadWriteSegment(basePath, 0, 1024, 0, nil)
+	assert.NoError(t, err)
+	for i := int64(0); i < 51; i++ {
+		err := rw.Append(i, fmt.Appendf(nil, "entry-%d", i))
+		assert.NoError(t, err)
+	}
+
+	err = rw.Append(51, fmt.Appendf(nil, "entry-%d", 51))
+	assert.Error(t, err)
```
merlimat: We should do:

```suggestion
	assert.ErrorIs(t, err, ErrSegmentFull)
```

merlimat (approved):

PR #628, opened 2025-02-26
fix(620): wait shard controller loop exit when close it

lsytj0413:
Fix: https://github.com/streamnative/oxia/issues/620


lsytj0413:
Ref: https://github.com/streamnative/oxia/pull/622

merlimat (approved):

PR #634, opened 2025-03-04
feat(631): add wal benchmark test

lsytj0413:
Fix: https://github.com/streamnative/oxia/issues/631


mattisonchao:
Hi, @lsytj0413 

Because the CI is not suitable for standard performance testing(the resource is not stable), would you mind making it a CLI? we can create a stable cluster to test it nightly then. :)

lsytj0413:
> Hi, @lsytj0413
> 
> Because the CI is not suitable for standard performance testing(the resource is not stable), would you mind making it a CLI? we can create a stable cluster to test it nightly then. :)
> 
> For example, we can pass different sizes of data as parameters to test the performance.

Ok, i will work on it.

merlimat:
The `Benchmark_` tests are only executed if you do `go benchmark ...` not in the regular tests.

merlimat (approved):

mattisonchao:
> The Benchmark_ tests are only executed if you do go benchmark ... not in the regular tests.

Yes, even that. we'd better test different sizes by parameters.

PR #644, opened 2025-03-20
fix: fallback to normal delete when deleting a small ranges

mattisonchao:
### Motivation

In the current implementation, we will get terrible performance for iterator-related operations when we have many small `DeleteRange` ranges.  
FYI: https://rocksdb.org/blog/2018/11/21/delete-range.html

It will make our `putWithSequence` method execute very slowly. (based on the current memtable size)

The reason is that the implementation of pebble(rocksDB) will create the fragmented range deletion tombstones for range deletion. Even Pebble will cache the tombstones, but it will invalidate when you get a new delete range operation. Therefore, we need to rebuild the deletion iterator after the new `DeleteRange` operation. And it will get worse along with more `DeleteRange` operations.
FYI:  https://github.com/cockroachdb/pebble/blob/dbc1c128682f7efcdb76352432249780b12447f7/mem_table.go#L248-L251

Also, rocksdb [article](https://rocksdb.org/blog/2018/11/21/delete-range.html) also mentioned that in `Future Work` section. 

### Modification

- To support atomic operation for the user side, we can fallback the `DeleteRange` operation back to the normal delete if the actual keys are lower than `DeleteRangeThreshold(default 100)`.

### Next

1. Introduce policies to make `DeleteRangeThreshold` configurable. 
2. Introduce policies to configure the memtable size. because the smaller size can trigger flush and compact the deletion tombstones.
3. other small improvements.


server/kv/db_benchmark_deleterange_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+	"time"
+
+	"github.com/streamnative/oxia/common"
+	"github.com/streamnative/oxia/proto"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/utils/pointer"
+)
+
+func BenchmarkGenerate100(b *testing.B) {
```
mattisonchao: This is the benchmark test. if you want to understand more details, you can check out to commit 8996085ca88f2ae394213df55595acdc16cff86f and check the `tests/deleterange_test.go` . that is using the actual data.

hangc0276:
@mattisonchao In BookKeeper, we also use deleteRange to delete keys, but we found RocksDB background compaction job will skip compacting the SST files whose key are deleted by deleteRange operation. https://github.com/apache/bookkeeper/pull/4555

Would you please double check if the SST file are compacted?

server/kv/db_benchmark_deleterange_test.go
merlimat: Do you have the before vs after results?

mattisonchao:
> In BookKeeper, we also use deleteRange to delete keys, but we found RocksDB background compaction job will skip compacting the SST files whose key are deleted by deleteRange operation. https://github.com/apache/bookkeeper/pull/4555

> Would you please double check if the SST file are compacted?

you are mentioning the disk size issue. the current problem is performance issue.

server/kv/db_benchmark_deleterange_test.go
mattisonchao: from benchmark test
```
main branch

goos: darwin
goarch: arm64
pkg: github.com/streamnative/oxia/server/kv
cpu: Apple M1 Pro
BenchmarkDeleteRange
BenchmarkDeleteRange-10    	    8863	  34814860 ns/op
PASS

this branch

goos: darwin
goarch: arm64
pkg: github.com/streamnative/oxia/server/kv
cpu: Apple M1 Pro
BenchmarkDeleteRange
BenchmarkDeleteRange-10    	   10000	    848611 ns/op
PASS


```

hangc0276:
> > In BookKeeper, we also use deleteRange to delete keys, but we found RocksDB background compaction job will skip compacting the SST files whose key are deleted by deleteRange operation. [apache/bookkeeper#4555](https://github.com/apache/bookkeeper/pull/4555)
> 
> > Would you please double check if the SST file are compacted?
> 
> you are mentioning the disk size issue. the current problem is read performance issue.

@mattisonchao It will also impact the read performance

mattisonchao:
> It will also impact the read performance

we are using in-memory implementation for benchmark. and also the data didn't exceed the memtable size.
but I would like to understand why it will also impact the read performance.

merlimat:
> but I would like to understand why it will also impact the read performance.

In an LSM you start from the top and you go down the tree of SST files. With point `get()` request is usually fine: If you find a `delete-rage` tombstone in the tree exploration, you can say that the key does not exist anymore. 

The problem is typically around handling iterators, because you have to remember the tombstones in the levels above the one you're currently exploring.

mattisonchao:
> In an LSM you start from the top and you go down the tree of SST files. With point get() request is usually fine: If you find a delete-rage tombstone in the tree exploration, you can say that the key does not exist anymore.

Yes.

> The problem is typically around handling iterators, because you have to remember the tombstones in the levels above the one you're currently exploring.

Yes, and it will construct a skyline to improve the read performance. but it's expensive to construct(maybe Pebble did cache for sst).  therefore, we should reduce the small range `DeleteRange` as much as possible. :)

server/kv/db.go
```diff
@@ -596,17 +596,41 @@ func (d *db) applyDelete(batch WriteBatch, notifications *notifications, delReq
...
 
-	if err := updateOperationCallback.OnDeleteRange(batch, delReq.StartInclusive, delReq.EndExclusive); err != nil {
+	it, err := batch.RangeScan(delReq.StartInclusive, delReq.EndExclusive)
+	if err != nil {
 		return nil, err
 	}
-
-	if err := batch.DeleteRange(delReq.StartInclusive, delReq.EndExclusive); err != nil {
-		return nil, errors.Wrap(err, "oxia db: failed to delete range")
+	var validKeys []string
```
merlimat, line 610: I'd try to find a way to avoid building this key array, since it could be very big.

One strategy could be:
 1. start doing point delete operations
 2. when we cross the threshold, we continue with the delete range

merlimat (approved):

PR #654, opened 2025-03-26
fix: fix similar issue as #653

mattisonchao:
### Motivation

similar as #653. this PR is fixing the same issue for the List interface. and add fast returns for other endpoints to avoid any potential issues in the future.

### Modification

- Return missing error


codelipenghui (approved):

codelipenghui:
I have verified with chaos testing, https://github.com/streamnative/oxia/pull/653 and this PR can completely resolve the topic load timeout issue.

Issue #661, opened 2025-04-09
Bug: too many outstanding goroutines


mattisonchao:
### Motivation

[goroutine.txt](https://github.com/user-attachments/files/19669584/goroutine.txt)

PR #662, opened 2025-04-09
feat: introduce stream callback

mattisonchao:
### Motivation

We've met several times on channel-related issues. 

- https://github.com/streamnative/oxia/pull/576
- https://github.com/streamnative/oxia/pull/652
- https://github.com/streamnative/oxia/issues/661
- ...

This PR is trying to introduce a stream callback to avoid managing the complex lifecycle of the channel. also, abstract batching logic can make it testable. 

### Modification

- Support `StreamCallback` 
- Support `BatchStreamOnce`
- Adding some tests


common/callback/callback.go
```diff
@@ -36,3 +36,21 @@ type Callback[T any] interface {
...
+	// OnNext is called whenever a new data item of type T is received from the stream.
+	// It processes the received data item and returns an error if any issues occur during processing.
+	// This method allows for custom logic to be applied to each individual data item in the stream.
+	OnNext(t T) error
+
+	// Complete is called when the stream has ended, either successfully or due to an error.
+	// The 'err' parameter indicates the status of the stream completion.
+	// If the stream ended successfully, 'err' will be nil. Otherwise,
+	// it will contain the error that caused the stream to terminate.
+	Complete(err error)
```
merlimat: Should we use `OnComplete(err error)` here to maintain consistency?

common/callback/once.go
```diff
@@ -61,3 +61,51 @@ func NewOnce[T any](onComplete func(t T), onError func(err error)) Callback[T] {
...
+// Complete is a method of the Stream struct. It is called when the stream is completed either successfully or due to an error.
+func (s *Stream[T]) Complete(err error) {
+	if !s.completed.CompareAndSwap(false, true) {
+		return
+	}
+	s.onComplete(err)
+}
+
+// NewStreamOnce is a factory function used to create a new Stream instance.
+func NewStreamOnce[T any](onNext func(t T), onComplete func(err error)) *Stream[T] {
```
merlimat: Why 2 functions instead of taking a `StreamCallback` interface here? Does it make it easier to invoke?

common/callback/callback.go
```diff
@@ -36,3 +36,21 @@ type Callback[T any] interface {
 	// CompleteError is invoked when the operation fails, providing an error 'err' indicating the failure reason.
 	CompleteError(err error)
 }
+
+// StreamCallback is a generic interface designed to handle streaming data.
+// It provides methods to process each incoming data item and to signal the completion of the stream,
+// which can be used in scenarios where data is processed in a streaming fashion,
+// such as handling network streams or reading large files in chunks.
+type StreamCallback[T any] interface {
```
merlimat, line 40: Would it make sense to also have a `BatchedStreamCallback` interface?

common/callback/batch_once.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// completed is an atomic boolean value used to mark whether the processing is completed.
+	// Using atomic operations ensures thread - safety in a concurrent environment.
+	completed atomic.Bool
+	// onFlush is a callback function that will be called when the batch - processing conditions are met.
+	// It takes a slice containing the data elements to be processed and returns an error if any.
+	onFlush func(container []T) error
+	// onFinish is a callback function that will be called when the processing is completed or an error occurs.
+	// It takes an error parameter to indicate the status of the processing.
+	onFinish func(err error)
```
merlimat: Maybe `onComplete` to keep consistency.

common/callback/batch_once.go
```diff
@@ -0,0 +1,89 @@
...
+// Complete is called when the stream processing is completed successfully.
+func (b *BatchStreamOnce[T]) Complete(err error) {
+	if !b.completed.CompareAndSwap(false, true) {
+		return
+	}
+	if err != nil {
+		b.onFinish(err)
+		return
+	}
+	b.onFinish(b.onFlush(b.container))
```
merlimat: if the list is empty, we should probably skip the flush

common/callback/callback.go
mattisonchao: applied.

common/callback/once.go
mattisonchao: using the interface for it now

common/callback/callback.go
mattisonchao: ` BatchedStreamCallback` just an implementation of `StreamCallback`.

common/callback/batch_once.go
mattisonchao: applied.

common/callback/batch_once.go
mattisonchao: fixed.

merlimat (approved):

PR #663, opened 2025-04-11
feat: introduce checkpoint to detect and auto discard dirty data

mattisonchao:
### Motivation

Because of metadata requirements, we need to generate some internal keys and version IDs in Oxia. However, we might introduce unexpected dirty data by introducing some new features or changes that change the state machine(oxia server). Especially by rolling upgrades, the leader and followers might use different logic. 

This PR Is introducing a checkpoint mechanism which can detect and auto-discard dirty replicas. The logic is as follows:

image

The leader will piggyback checkpoint messages with the same internal states(commit-offset, version-id). and then replicate data to the follower. when the replication starts, the leader will read the last checkpoint from the database and send it to followers as `termCheckpoint` and then the followers can use this as the expectation for further validation. 

we can configure some policies to handle the failure once we find the inconsistency.

- Warn: the follower will print error logs to notify admin about the data inconsistency.
- Discard: the follower will return an `ErrUnmatchedCheckpoint` error and notify the leader to send a new snapshot.

### Modification

- Introduce namespace level policies.
- Introduce checkpoint policy.
- Support checkpoint validation logic on both leader and follower.


mattisonchao:
Let us use this thread for more discussion on the data consistency validation.

PR #683, opened 2025-05-14
Added `oxia wal scan` tool to analyize WAL

merlimat:


cmd/wal/scan/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+		SyncData:    false,
+	})
+
+	wal, err := factory.NewWal(common.WalOption.Namespace, common.WalOption.Shard, nil)
+	if err != nil {
+		return err
+	}
+	defer wal.Close()
+
+	reader, err := wal.NewReader(wal.FirstOffset() - 1)
```
mattisonchao: Let's give user a option to set start offset.  :)

cmd/wal/scan/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+		if err != nil {
+			return err
+		}
+
+		lev := proto.LogEntryValue{}
+		if err = lev.UnmarshalVTUnsafe(le.Value); err != nil {
+			return err
+		}
+
+		if options.json {
```
mattisonchao, line 105: To simplify, we can support JSON only, which can be easily processed by jq. :)

merlimat: It's actually a lot of info. It's quite more convenient the one line per each operation

mattisonchao: okay.

cmd/wal/scan/cmd.go
merlimat: Done

mattisonchao (approved):

PR #686, opened 2025-05-15
Support Get() operation using secondary index

merlimat:
Allow to specify a `UseIndex()` option on `Get()` operations. 

The key is then compared with the selected secondary index. It works also with `ComparisonLower()`, `ComparisonHigher()` etc..


server/secondary_indexes.go
```diff
@@ -153,11 +155,22 @@ func secondaryIndexKey(primaryKey string, si *proto.SecondaryIndex) string {
...
 	}
 
-	return url.PathUnescape(matches[1])
+	return url.PathUnescape(matches[2])
+}
+
+func secondaryIndexPrimaryAndSecondaryKey(completeKey string) (primaryKey string, secondaryKey string, err error) {
+	matches := secondaryIdxFormatRegex.FindStringSubmatch(completeKey)
+	if len(matches) != 3 {
+		return "", "", nil
```
mattisonchao: it should be an error. :/

server/secondary_indexes.go
```diff
@@ -277,3 +302,92 @@ func (it *secondaryIndexRangeIterator) Value() (*proto.GetResponse, error) {
 
 	return gr, err
 }
+
+// /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
mattisonchao, line 308: why do we need this?   :)

server/secondary_indexes.go
merlimat: Good catch. Fixed

server/secondary_indexes.go
merlimat: These are just to visually separate different sections

mattisonchao (approved):

PR #733, opened 2025-07-03
feat(coordinator): support node controller graceful close

mattisonchao:
### Motivation

Support node controller graceful close.

### Modification

- Move background ping to an extra goroutine.
- Wait until all the background goroutines finish when closing.

### DIFF
--- a/coordinator/controllers/node_controller.go
+++ b/coordinator/controllers/node_controller.go
@@ -108,18 +108,13 @@ func (n *nodeController) SetStatus(status NodeStatus) {
 func (n *nodeController) maybeInitHealthClient() {
 	n.healthClientOnce.Do(func() {
 		_ = backoff.RetryNotify(func() error {
-			select {
-			case <-n.ctx.Done():
-				return nil
-			default:
-				health, closer, err := n.rpc.GetHealthClient(n.node)
-				if err != nil {
-					return err
-				}
-				n.healthClient = health
-				n.healthClientCloser = closer
-				return nil
+			health, closer, err := n.rpc.GetHealthClient(n.node)
+			if err != nil {
+				return err
 			}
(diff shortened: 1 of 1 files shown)


coordinator/controllers/node_controller.go
```diff
@@ -62,316 +65,307 @@ type NodeController interface {
...
-			n.healthCheckLoop(ctx, health)
-		},
-	)
-
-	watch, err := health.Watch(ctx, &grpc_health_v1.HealthCheckRequest{Service: ""})
-	if err != nil {
-		return err
+	n.statusLock.Lock()
+	if n.status != Running && n.status != Draining { // double check
+		return
```
RobertIndie, line 260: The statusLock won't be unlocked here.

coordinator/controllers/node_controller.go
```diff
@@ -62,316 +65,307 @@ type NodeController interface {
...
-		func() {
-			defer cancel()
-			n.healthCheckLoop(ctx, health)
-		},
-	)
-
-	watch, err := health.Watch(ctx, &grpc_health_v1.HealthCheckRequest{Service: ""})
-	if err != nil {
-		return err
+	n.statusLock.Lock()
```
RobertIndie, line 257: There is already a lock operation in n.Status. I don't think we need double lock here.

RobertIndie (commented):
Left some comments

coordinator/controllers/node_controller.go
mattisonchao: we need it because we have a mutable operation after. :)

coordinator/controllers/node_controller.go
mattisonchao: fixed.

PR #739, opened 2025-08-13
feat: support follower read

mattisonchao:
### Motivation

Support experimental follower read to help the admin check the cluster consistency and status. 

### Modification

- Introduce `ReadableController` as an interface and reuse the read-related logic on both leader and follower.
- Expose read-related logic for `FollowerController`
- Support `EventualConsistency` for the Golang client and CMD.


cmd/client/get/cmd.go
```diff
@@ -30,12 +30,13 @@ var (
...
-	partitionKey   string
-	comparisonType string
-	secondaryIndex string
+	key              string
+	hexDump          bool
+	includeVersion   bool
+	partitionKey     string
+	comparisonType   string
+	secondaryIndex   string
+	consistencyLevel string
```
merlimat: I think we should be using an enum. It should work with Cobra as well

oxia/options_base.go
```diff
@@ -80,3 +82,41 @@ func PartitionKey(partitionKey string) BaseOption {
 		partitionKey: &partitionKey,
 	}
 }
+
+// --------------------------------------------------------------------------------------------
+
+type ConsistencyOpt struct {
```
merlimat: I wouldn't leave this as a per-operation option. I'd prefer to have it as a client instance wide option.

mattisonchao: changed it.

cmd/client/get/cmd.go
mattisonchao: moved to client options as a bool flag.

PR #742, opened 2025-08-26
fix(server): fix the ephemeral key leak by non-atomic operation

mattisonchao:
### Motivation

A non-atomic operation will leak some ephemeral keys if the new KV is put after the session list.

### Modification

- Make the session index/ephemeral keys cleanup become an atomic operation.

### DIFF
--- a/server/session_manager.go
+++ b/server/session_manager.go
@@ -20,23 +20,22 @@ import (
 	"io"
 	"log/slog"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
-	"go.uber.org/multierr"
-
-	"github.com/oxia-db/oxia/common/constant"
-
 	"github.com/oxia-db/oxia/common/collection"
+	"github.com/oxia-db/oxia/common/constant"
 	"github.com/oxia-db/oxia/common/metric"
 	"github.com/oxia-db/oxia/proto"
 	"github.com/oxia-db/oxia/server/kv"
+	"github.com/pkg/errors"
 )
 
 const (
 	sessionKeyPrefix = constant.InternalKeyPrefix + "session"
 	sessionKeyFormat = sessionKeyPrefix + "/%016x"
+	sessionKeyLength = len(sessionKeyPrefix) + 1 + 16
 )
 
 type SessionId int64
@@ -61,6 +60,16 @@ func KeyToId(key string) (SessionId, error) {
 	}
 
 	return SessionId(id), nil
+}
+
+func IsSessionKey(key string) bool {
+	if !strings.HasPrefix(key, sessionKeyPrefix) {
+		return false
+	}
+	if len(key) != sessionKeyLength {
+		return false
+	}
+	return true
 }
 
 // --- SessionManager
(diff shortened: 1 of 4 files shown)


oxia/options_client.go
```diff
@@ -213,3 +219,12 @@ func WithAuthentication(authentication auth.Authentication) ClientOption {
 		return options, nil
 	})
 }
+
+// WithSessionKeepAliveTicker is an internal API used to control the duration
+// of the session keep-alive ticker. This is for experimental use only.
+func WithSessionKeepAliveTicker(ticker time.Duration) ClientOption {
```
merlimat: I wouldn't expose this in API (even if comment says it's internal). We should be able to modify the value by accessing the private variable in the tests

server/session_manager.go
```diff
@@ -352,52 +362,94 @@ func deleteShadow(batch kv.WriteBatch, key string, existingEntry *proto.StorageE
...
 }
 
-func (*sessionManagerUpdateOperationCallbackS) OnDeleteWithEntry(batch kv.WriteBatch, key string, value *proto.StorageEntry) error {
-	_, err := deleteShadow(batch, key, value)
-	return err
+func (*sessionManagerUpdateOperationCallbackS) OnDeleteWithEntry(batch kv.WriteBatch, notification *kv.Notifications, key string, entry *proto.StorageEntry) error { //nolint:revive
+	if _, err := deleteShadow(batch, notification, key, entry); err != nil {
+		return err
+	}
+	if IsSessionKey(key) {
```
merlimat: nit: 

```
if !IsSessionKey(key) {
    return nil;
}
....
```

oxia/options_client.go
mattisonchao: moved the test into the oxia package and made the `WithSessionKeepAliveTicker` private.

server/session_manager.go
mattisonchao: fixed.

server/session_manager.go
```diff
@@ -352,52 +362,95 @@ func deleteShadow(batch kv.WriteBatch, key string, existingEntry *proto.StorageE
...
 }
 
-func (*sessionManagerUpdateOperationCallbackS) OnDelete(batch kv.WriteBatch, key string) error {
+func (s *sessionManagerUpdateOperationCallbackS) OnDelete(batch kv.WriteBatch, notification *kv.Notifications, key string) error {
 	se, err := kv.GetStorageEntry(batch, key)
-	defer se.ReturnToVTPool()
-
-	if errors.Is(err, kv.ErrKeyNotFound) {
-		return nil
+	if err != nil {
```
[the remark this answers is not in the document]
mattisonchao: fixed.

server/session_manager.go
```diff
@@ -352,52 +362,89 @@ func deleteShadow(batch kv.WriteBatch, key string, existingEntry *proto.StorageE
...
+		escapedEphemeralKey := it.Key()
+		// delete the ephemeral key index
+		if err := batch.Delete(escapedEphemeralKey); err != nil {
+			return err
+		}
+		unescapedEphemeralKey, err := url.PathUnescape(escapedEphemeralKey[len(sessionKey)+1:])
+		if err != nil {
+			return err
+		}
+		if unescapedEphemeralKey != "" {
```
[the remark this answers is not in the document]
mattisonchao: No, we are using range-scan to get the key.

merlimat (approved):

PR #609, opened 2025-02-19
test(506): regarded as a successful closure  when session not found

lsytj0413:
Fix：https://github.com/streamnative/oxia/issues/506

PR #744, opened 2025-08-28
fix(session): avoid complete rpc without response

mattisonchao:
### Motivation

fixes: https://github.com/oxia-db/oxia/issues/745

The current logic will return `nil,nil` to the client when the session is not found. which is confusing the client side because we didn't give the correct response body, but an OK status.

### DIFF
--- a/server/public_rpc_server.go
+++ b/server/public_rpc_server.go
@@ -421,8 +421,8 @@ func (s *publicRpcServer) CloseSession(ctx context.Context, req *proto.CloseSess
 			)
 			return nil, err
 		}
-
 		s.log.Warn("Session not found, it should already closed")
+		return &proto.CloseSessionResponse{}, nil
 	}
 	return res, nil
 }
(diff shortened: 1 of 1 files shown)


server/public_rpc_server.go
```diff
@@ -415,14 +415,9 @@ func (s *publicRpcServer) CloseSession(ctx context.Context, req *proto.CloseSess
...
-			s.log.Warn(
-				"Failed to close session",
-				slog.Any("error", err),
-			)
-			return nil, err
+			s.log.Warn("Failed to close session", slog.Any("error", err))
 		}
-
-		s.log.Warn("Session not found, it should already closed")
+		return nil, err
```
merlimat, line 420: This is going to propagate the `SessionNotFound` back to client, right?

mattisonchao: Yes, I was considering hiding it. But on the server-side, we should give a clear error to the client and then let the client decide what they should do. :)

mattisonchao: Plus, we propagated the error to the client at the beginning of the system. 

#609 changed the behaviour.  (I accept we don't print the warn log on the server, but we shouldn't hide anything )

merlimat: 👍  Agree.

merlimat (approved):

PR #753, opened 2025-10-10
[feature] Add coordinator HTTP Metadata support

dao-jun:
Add HTTP coordinator Metadata support.


coordinator/metadata/metadata_http.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	c := &Container{Version: expectedVersion, ClusterStatus: cs}
+	bts, err := json.Marshal(c)
+	if err != nil {
+		slog.Error("Marshall failed", slog.Any("c", c), slog.Any("error", err))
+		return "", err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, m.metadataApiURL, bytes.NewBuffer(bts))
```
merlimat, line 101: The version control here is not atomic. It can definitely lead to lost updates

PR #757, opened 2025-10-15
Raft based metadata provider

merlimat:
Example of usage: 

```
bin/oxia coordinator --conf test.yaml \
                --metadata raft \
                --raft-address node-1:9000 \
                --raft-bootstrap-nodes node-1:9000,node-2:9000,node-3:9000
```

### DIFF
--- a/common/logging/logger.go
+++ b/common/logging/logger.go
@@ -77,12 +77,22 @@ func ConfigureLogger() {
 		Logger()
 
 	if !LogJSON {
-		writer := zerolog.SyncWriter(zerolog.ConsoleWriter{
-			Out:        os.Stdout,
-			TimeFormat: time.StampMicro,
-		})
+		if raceEnabled {
+			// Use simple non-colored logger in tests to avoid
+			// data races
+			zerologLogger = zerolog.New(os.Stdout).
+				With().
+				Timestamp().
+				Stack().
+				Logger()
+		} else {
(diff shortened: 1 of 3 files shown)


coordinator/metadata/metadata_raft.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Create snapshot store
+	snapshotStore, err := raft.NewFileSnapshotStore(dataDir, 2, os.Stderr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create snapshot store")
+	}
+
+	// Create BoltDB-based stable store and log store
+	boltDBPath := filepath.Join(dataDir, "raft.db")
+	store, err := boltdb.NewBoltStore(boltDBPath)
```
mattisonchao: I understand the boltDB will be more suitable for the Raft case. 

But I am still considering if it will make the background knowledge requirement more complex for Oxia?

mattisonchao (approved):

coordinator/metadata/metadata_raft.go
merlimat: Good point. Let me check how difficult it would be to have it based on the KV interface we have already.

merlimat:
There are problems with the deps that are making Pebble fail in some weird ways (that's also the reason why I was trying with #760)

PR #760, opened 2025-10-15
Upgrade to Pebble 2.1.0

merlimat:
Upgrade & downgrade work seamlessly since we're pinning  the disk format version.


merlimat:
This is blocked by this problem: https://github.com/cockroachdb/pebble/issues/5436

PR #762, opened 2025-10-15
Use original context if available in Timer.Done()

merlimat:


common/metric/latency_histogram.go
```diff
@@ -31,7 +31,11 @@ type Timer struct {
 }
 
 func (tm Timer) Done() {
-	tm.histo.histo.Record(context.Background(), float64(time.Since(tm.start).Microseconds())/1000.0, tm.histo.attrs)
+	tm.DoneCtx(context.Background())
+}
+
+func (tm Timer) DoneCtx(ctx context.Context) {
```
mattisonchao, line 37: If the context is indeed useful, we could change the method signature directly. :) 

```
func (tm Timer) Done(context.Context) {
```

merlimat: Not sure it's super useful, though it was messing up with the linter. It was getting different linter comments on Mac vs CI workflow...

PR #786, opened 2025-10-24
Handle deletion of swapped out node in a background periodic task

merlimat:
https://github.com/oxia-db/oxia/issues/845

After a `SwapNode()` initiated leader election, let's not wait until the removed node has deleted the shard since this is delaying the leader election time.

Instead, move the logic out to a periodic task with retries.

Note: the node is immediately moved to a different category in the status and marked in the `PendingDeleteShardNodes` list. This is because it will not participate anymore in the newTermQuorum.

### DIFF
--- a/coordinator/controllers/mock_test.go
+++ b/coordinator/controllers/mock_test.go
@@ -130,6 +130,7 @@ type mockPerNodeChannels struct {
 }
 
 const defaultTimeout = 10 * time.Second
+const defaultNegativeTimeout = 1 * time.Second
 
 func (m *mockPerNodeChannels) expectBecomeLeaderRequest(t *testing.T, shard int64, term int64, replicationFactor uint32) {
 	t.Helper()
@@ -159,6 +160,17 @@ func (m *mockPerNodeChannels) expectNewTermRequest(t *testing.T, shard int64, te
 	assert.Equal(t, shard, r.Shard)
 	assert.Equal(t, term, r.Term)
 	assert.Equal(t, notificationsEnabled, r.Options.EnableNotifications)
+}
+
+func (m *mockPerNodeChannels) expectNoMoreNewTermRequest(t *testing.T) {
+	t.Helper()
+
+	select {
+	case <-m.newTermRequests:
+		assert.Fail(t, "should not have received any new term request")
+	case <-time.After(defaultNegativeTimeout):
+		// expected
+	}
 }
 
 func (m *mockPerNodeChannels) expectDeleteShardRequest(t *testing.T, shard int64, term int64) {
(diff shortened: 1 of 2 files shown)


coordinator/controllers/shard_controller.go
```diff
@@ -55,6 +55,8 @@ const (
 	catchupTimeout = 5 * time.Minute
 
 	chanBufferSize = 100
+
+	DefaultPeriodicTasksInterval = 1 * time.Minute
```
mattisonchao, line 59: Maybe we could use backoff to avoid being delayed too much? 

FYI: https://github.com/oxia-db/oxia/pull/783/files#diff-9d63c1fdefc96b624e16c502d1a7eaeb5120498d2a50422d4bcb7f05d5574144R361

coordinator/controllers/shard_controller.go
```diff
@@ -229,6 +235,8 @@ func (s *shardController) run() {
 		slog.Any("leader", s.shardMetadata.Leader),
 	)
 
+	periodicTasksTimer := time.NewTicker(s.periodicTasksInterval)
```
mattisonchao, line 238: It's better to stop it in defer. :)

coordinator/controllers/shard_controller.go
```diff
@@ -390,21 +401,15 @@ func (s *shardController) electLeader() (string, error) {
 	}
 
 	if err = s.becomeLeader(newLeader, followers); err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "failed to become leader for node %s", newLeader.GetIdentifier())
 	}
 
 	metadata := s.shardMetadata.Clone()
 	metadata.Status = model.ShardStatusSteadyState
+	metadata.PendingDeleteShardNodes = metadata.RemovedNodes
```
mattisonchao: Should we do `append` here?

coordinator/controllers/shard_controller.go
merlimat: It's not needed since Go 1.23 :)

coordinator/controllers/shard_controller.go
merlimat: Yes, good catch!

coordinator/controllers/shard_controller.go
merlimat: I think delaying the shard deletion by 1 min is ok and should never be a problem. If any backoff, I'd do a backoff from 1 min -> 1h -> 1day for this kind of operations.

mattisonchao: NO, we shouldn't do a 1-day backoff. The shard deletion will be delayed for a very long time. :)

PR #788, opened 2025-10-24
fixes: deadlock when swap `waitForFollowersToCatchUp`

mattisonchao:
### Motivation

fixes: https://github.com/oxia-db/oxia/issues/845

In the current implementation. The shard controller might be stuck in the `swap` method due to `waitForFollowersToCatchUp`.  

The root cause is the swap waiting for `waitForFollowersToCatchUp`, but the new follower needs to wait for `newTermAndAddFollowerOp` to be fenced and added as a follower.

### Modification

- Move `waitForFollowersToCatchUp` operation out of `shardController` main loop.

### DIFF
--- a/coordinator/controllers/shard_controller_test.go
+++ b/coordinator/controllers/shard_controller_test.go
@@ -204,6 +204,10 @@ func TestShardController_StartingWithLeaderAlreadyPresent(t *testing.T) {
 
 	n1.expectGetStatusRequest(t, shard)
 	n1.GetStatusResponse(1, proto.ServingStatus_LEADER, 0, 0)
+	n2.expectGetStatusRequest(t, shard)
+	n2.GetStatusResponse(1, proto.ServingStatus_FOLLOWER, 0, 0)
+	n3.expectGetStatusRequest(t, shard)
+	n3.GetStatusResponse(1, proto.ServingStatus_FOLLOWER, 0, 0)
 
 	n1.expectNoMoreNewTermRequest(t)
 	n2.expectNoMoreNewTermRequest(t)


coordinator/controllers/shard_controller.go
```diff
@@ -422,7 +411,15 @@ func (s *shardController) electLeader() (string, error) {
...
 	}
 
-	s.keepFencingFailedFollowers(followers)
+	go process.DoWithLabels(
+		s.currentElectionCtx,
+		map[string]string{
+			"oxia":  "election-fencing-failed-followers",
+			"shard": fmt.Sprintf("%d", s.shard),
+		}, func() {
+			s.keepFencingFailedFollowers(s.shardMetadata.Term, s.shardMetadata.Ensemble, s.shardMetadata.Leader, followers)
```
merlimat: We are reading the metadata from the go-routine, outside of its mutex. We can make a clone before or acquire the mutex here

coordinator/controllers/shard_controller.go
```diff
@@ -464,111 +461,89 @@ func (s *shardController) deletingRemovedNodes() error {
...
+			slog.Int64("term", term),
 		)
 		return
 	}
-
 	// Identify failed followers
-	for _, sa := range s.shardMetadata.Ensemble {
-		if sa == *s.shardMetadata.Leader {
+	group := sync.WaitGroup{}
+	defer group.Wait()
```
merlimat: What's the reason for waiting at the end of the method, if we are not going to return any error?

coordinator/controllers/shard_controller.go
mattisonchao: fixed.

coordinator/controllers/shard_controller.go
mattisonchao: It's useless so far. Let me remove it. :)

mattisonchao:
rely on the #790

PR #791, opened 2025-10-25
refactor: introduce thread safe shard metadata

mattisonchao:
### Motivation

This PR introduces a thread-safe metadata structure to hide the locks behind operations. It also changes the way election pipeline metadata is accessed to use mutable clone metadata to avoid touching the metadata structure with a lock.

### DIFF
--- a/coordinator/controllers/shard_controller_metadata.go
+++ b/coordinator/controllers/shard_controller_metadata.go
@@ -1,3 +1,17 @@
+// Copyright 2023-2025 The Oxia Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


coordinator/controllers/shard_controller.go
```diff
@@ -59,6 +59,58 @@ const (
 	DefaultPeriodicTasksInterval = 1 * time.Minute
 )
 
+// --------------------------- Metadata -----------------------------
```
merlimat: Maybe move to a separate file? Also, maybe use an interface

merlimat (approved):

coordinator/controllers/shard_controller.go
mattisonchao: moved to a separate file. 

> Also, maybe use an interface

It's only the shard controller memory metadata. We don't need to give it an interface so far. We could add it when we really need it. :)

PR #792, opened 2025-10-25
[feat][admin] Initial Oxia admin commit and add list-namespaces command.

dao-jun:
Add oxia admin ist-namespaces command.

### DIFF
--- a/cmd/admin/listnamespaces/cmd.go
+++ b/cmd/admin/listnamespaces/cmd.go
@@ -15,8 +15,9 @@
 package listnamespaces
 
 import (
-	"github.com/oxia-db/oxia/cmd/admin/commons"
 	"github.com/spf13/cobra"
+
+	"github.com/oxia-db/oxia/cmd/admin/commons"
 
 	cc "github.com/oxia-db/oxia/cmd/client/common"
 	"github.com/oxia-db/oxia/oxia"


cmd/admin/listnamespaces/cmd.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	result := client.ListNamespaces()
+	if result.Error != nil {
+		return result.Error
+	}
+	cc.WriteOutput(cmd.OutOrStdout(), result.Namespaces)
```
mattisonchao, line 52: for admin, we might need to consider adding JSON or other format to help CI. :)

common/constant/constants.go
```diff
@@ -24,6 +24,7 @@ const (
 
 	DefaultPublicPort   = 6648
 	DefaultInternalPort = 6649
+	DefaultAdminPort    = 6651
```
mattisonchao, line 27: curious why you choose 6651? not 6650?

mattisonchao (approved):

common/constant/constants.go
dao-jun: This is to distinguish it from Pulsar

cmd/admin/listnamespaces/cmd.go
dao-jun: In the future, for more complex commands, I will use JSON format. In the current simple command, I think text is already sufficient

common/constant/constants.go
mattisonchao: okay, but 6651 is recommended also for Pulsar with TLS. (although that's not default for pulsar `broker.conf`)  

FYI: https://pulsar.apache.org/docs/next/security-tls-transport/#step-2-configure-brokers

PR #794, opened 2025-10-27
[server][improve] Add WAL cache to optimize replication.

dao-jun:
Add WAL LogEntry cache to improve replication.
Bypass page-cache and eliminate deserialization overhead when tailing-read the WAL.

Under ideal conditions, the `oxia_server_wal_read_latency_milliseconds_sum` can be 0.
image

Test the WAL via wal-perf
before:
image
after:
image

The Read/Write throughput increase about **32%**

### DIFF
--- a/server/wal/wal.go
+++ b/server/wal/wal.go
@@ -116,7 +116,7 @@ type cacheValue struct {
 	length int
 }
 
-// LogEntryCache is a FIFO cache for log entries
+// LogEntryCache is a FIFO cache for log entries.
 type LogEntryCache struct {
 	sync.RWMutex
 	// Map to store offset to cacheValue mapping for O(1) lookup


server/wal/wal_impl.go
```diff
@@ -131,6 +134,12 @@ func newWal(namespace string, shard int64, options *FactoryOptions, commitOffset
 			"The time it takes to fsync the wal data on disk", labels),
 	}
 
+	if cache, err := lru.New(logEntryCacheSize); err == nil {
+		w.cacheHits = metric.NewCounter("oxia_server_wal_cache_hits",
```
merlimat: Should we also track the misses here?

server/wal/wal.go
```diff
@@ -24,6 +24,10 @@ import (
 	"github.com/oxia-db/oxia/proto"
 )
 
+const (
+	logEntryCacheSize int = 32
```
merlimat: It would be better if we could set `weighter` and use max number of bytes instead of entries here.

dao-jun: Resolved. But I don't understand what does `set weighter` mean, could you please explain in detail?

mattisonchao:
> I've considered this. Even in future single WAL instances, we still strive to distribute memory evenly among each shard. Otherwise, there will still be cache penetration, which is not much different from the current implementation

well... after deep thinking. I think the write cache could not help us more about this case.  because reader always happened after the data sync. If the write traffic is very large, the 2MiB buffer will never work as expected. 

Indeed, your benchmark proof some improvement, but that logic is different as oxia.  let me change some logics to make it match the implementation of Oxia. Plus, you could also use your implementation in the cluster benchmarking to see if any improvement on `oxia_server_wal_read_latency_milliseconds_sum `.

PR #798, opened 2025-10-29
fix: data lost by change ensemble

mattisonchao:
### Motivation

fixes: https://github.com/oxia-db/oxia/issues/845

We will update the memory metadata in the current implementation of change ensemble(swap node), then trigger the leader election, which will persist the new ensemble to metadata without any precondition. If there is any failure to update the elected, the ensemble will be changed and might be changed next time. 

Keeping moving shards without follower-caught-up validation will cause a data loss issue. 

### Modification

1.  Introduce the status of the follower caught up in the election.
2. Move the caught-up validation into election ownership, and check that the followers caught up after the new leader was elected. which can also be cancelled by `election#stop`.
3. Check the `IsReadyForChangeEnsemble` before changing the ensemble, and fail fast if the precondition is not satisfied.
4. Implement `copy-on-write` for `UpdateShardMetadata ` to avoid data racing.

### DIFF
--- a/cmd/admin/cmd.go
+++ b/cmd/admin/cmd.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/oxia-db/oxia/cmd/admin/commons"
+	"github.com/oxia-db/oxia/cmd/admin/listnodes"
 
 	"github.com/oxia-db/oxia/cmd/admin/listnamespaces"
 
@@ -39,4 +40,5 @@ func init() {
 	Cmd.PersistentFlags().StringVar(&commons.AdminConfig.AdminAddress, "admin-address", defaultAdminClientAddress, "Admin client address")
 
 	Cmd.AddCommand(listnamespaces.Cmd)
+	Cmd.AddCommand(listnodes.Cmd)
 }
--- a/cmd/admin/commons/mock_admin_client.go
+++ b/cmd/admin/commons/mock_admin_client.go
@@ -45,3 +45,13 @@ func (m *MockAdminClient) ListNamespaces() *oxia.ListNamespacesResult {
 		Error: errors.New("no namespaces available"),
 	}
 }
+
+func (m *MockAdminClient) ListNodes() *oxia.ListNodesResult {
+	args := m.MethodCalled("ListNodes")
+	if v, ok := args.Get(0).(*oxia.ListNodesResult); ok {
+		return v
+	}
+	return &oxia.ListNodesResult{
+		Error: errors.New("no namespaces available"),
+	}
+}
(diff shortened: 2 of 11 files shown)


dao-jun:
Can we just check if shardMeta.RemovedNodes is empty in shardController.swapNode? If it is, proceed with the swap; otherwise, skip it.

mattisonchao:
>Can we just check if shardMeta.RemovedNodes is empty in shardController.swapNode? If it is, proceed with the swap; otherwise, skip it.

Hi @dao-jun  

Good point, 

Unfortunately, the answer is no so far.  

The current `RemovedNodes` mechanism is only used to record node deletions, not to protect the quorum. The value of `RemovedNodes` is cleared after the election. However, to avoid data loss due to load balancer shard movement, we should prevent further ensemble changes before a new quorum is formed.

Plus, I mentioned "so far" because we still have another issue that related to failure recovery.  Once the issue is fixed, we can rely on the `RemovedNodes`.

coordinator/controllers/shard_controller_election.go
```diff
@@ -230,6 +238,55 @@ func (e *ShardElection) becomeLeader(term int64, leader model.Server, followers
...
+			e.Context,
+			map[string]string{
+				"oxia":      "shard-caught-up-monitor",
+				"namespace": e.namespace,
+				"shard":     fmt.Sprintf("%d", e.shard),
+				"node":      server.GetIdentifier(),
+			}, func() {
+				defer waitGroup.Done()
+				err := backoff.RetryNotify(func() error {
+					fs, err := e.provider.GetStatus(e.Context, server, &proto.GetStatusRequest{Shard: e.shard})
```
dao-jun, line 259: Wait until all followers caughtup with leader?

coordinator/controllers/shard_controller_election.go
```diff
@@ -362,6 +449,7 @@ func (e *ShardElection) start() (model.Server, error) {
 	term := mutShardMeta.Term
 	ensemble := mutShardMeta.Ensemble
 	leader := mutShardMeta.Leader
+	leaderEntry := candidatesStatus[*leader]
```
dao-jun, line 452: Did we ensure the coordinator receive newTermResponse from the current leader in `fenceNewTermQuorum`?

coordinator/controllers/shard_controller_election.go
```diff
@@ -387,9 +476,26 @@ func (e *ShardElection) start() (model.Server, error) {
...
 		)
 	})
+	e.Go(func() {
+		process.DoWithLabels(
+			e.Context,
+			map[string]string{
+				"oxia":  "election-monitor-followers-caught-up",
+				"shard": fmt.Sprintf("%d", e.shard),
+			}, func() {
+				e.ensureFollowerCaught(ensemble, leader, leaderEntry)
```
dao-jun, line 486: As we fenced the leader,  can the followers still caughtup with the leader?

coordinator/controllers/shard_controller_election.go
mattisonchao: yes.

merlimat (approved):

PR #800, opened 2025-11-03
Add listnodes admin tool

dao-jun:
Add listnodes admin tool

### DIFF
--- a/oxia/admin.go
+++ b/oxia/admin.go
@@ -33,6 +33,7 @@ type Node struct {
 	Name            *string
 	PublicAddress   string
 	InternalAddress string
+	Metadata        map[string]string
 }
 
 type ListNodesResult struct {
--- a/proto/admin.proto
+++ b/proto/admin.proto
@@ -44,7 +44,9 @@ message Node {
   optional string name = 1;
   string public_address = 2;
   string internal_address = 3;
+  map<string, string> metadata = 4;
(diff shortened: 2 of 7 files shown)


proto/admin.proto
```diff
@@ -26,11 +26,25 @@ service OxiaAdmin {
...
   repeated string namespaces = 1;
+}
+
+message ListNodesRequest {
+}
+
+message Node {
+  optional string name = 1;
+  string public_address = 2;
+  string internal_address = 3;
```
mattisonchao, line 46: maybe we can return node metadata along with node?

dao-jun: I will add `node-shards` to describe the shards which host on the node 
and add `describe-namespace` to describe the shard ensemble

oxia/admin.go
```diff
@@ -20,9 +20,23 @@ type AdminClient interface {
...
 }
+
+type Node struct {
+	Name            *string
+	PublicAddress   string
+	InternalAddress string
+	Metadata        map[string]string
+}
+
+type ListNodesResult struct {
```
mattisonchao, line 39: Why do we need this extra structure as a result? We could return `(Nodes, error)` directly.  :/

dao-jun: just follow oxia client's pattern

PR #836, opened 2025-12-12
feat: add CA certificate and IPv6 options to client configuration

ciiiii:
- Introduced `CACertFile` option for specifying a CA certificate file for TLS verification.
- Improve TLS verify to match target with both commmonName and DNSNames
- Added `DisableIPv6` option to allow users to disable IPv6 resolution, forcing the use of IPv4 addresses.
- Updated client pool and related tests to support the new options.

### DIFF
--- a/cmd/health/cmd_test.go
+++ b/cmd/health/cmd_test.go
@@ -26,7 +26,6 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/oxia-db/oxia/oxiad/common/rpc"
-
 	"github.com/oxia-db/oxia/oxiad/common/rpc/auth"
 )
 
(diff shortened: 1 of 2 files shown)


mattisonchao (commented):
## Review Summary

This PR adds `WithCACertFile` and `WithDisableIPv6` client options, along with a custom gRPC dialer for DNS resolution control, IPv4 preference, and enhanced TLS logging. CI is green.

---

### Critical

1. **`mergeCustomTLSConfig` silently drops most user TLS settings** — Only `ServerName`, `InsecureSkipVerify`, and `Certificates` are copied from the user-provided `*tls.Config`. All other fields (`CipherSuites`, `NextProtos`, `ClientAuth`, `CurvePreferences`, `VerifyConnection`, `VerifyPeerCertificate`, etc.) are discarded. This is a behavioral regression: previously `getTransportCredential` passed `cp.tls` directly to `credentials.NewTLS()`. Consider using `cp.tls.Clone()` as the base config and layering on hostname/RootCAs instead of rebuilding from scratch.

2. **`getTransportCredential` behavior change for `tls://` prefix** — Previously `tls://` without a custom config just did `credentials.NewTLS(&tls.Config{})`. Now it builds a full config with explicit system cert pool, `VerifyConnection` callback, etc. This silently changes TLS behavior for existing users.

### Important

3. **IPv4 always preferred over IPv6** — `sortIPs` unconditionally puts IPv4 before IPv6, going against RFC 6724 / Happy Eyeballs (RFC 8305). In IPv6-primary environments this adds latency from the failed IPv4 attempt. Consider making IPv4 preference opt-in rather than the default.

4. **No aggregate timeout for multi-IP dial** — Each IP gets `defaultDialTimeout` (5s). A host resolving to N IPs can block for N×5s. Consider an overall deadline from the context or parallel connection attempts.

5. **Unnecessary explicit system cert pool loading** — In `setRootCAs`, when no custom RootCAs are provided, Go's `crypto/tls` already uses the system pool by default when `RootCAs` is nil. Explicitly loading it adds overhead and loses lazy-loading.

### Minor

6. **Tests in wrong file** — New tests for `WithCACertFile`/`WithDisableIPv6` are in `options_delete_test.go`. Should be in `options_client_test.go` or similar.

7. **`GetHostname` exported unnecessarily** — Only used internally and in same-package tests. Exporting adds to public API surface without need.

8. **Health test changes appear unrelated** — Adding `--host=localhost` to health tests looks like a side-effect of the `passthrough:///` change. Worth calling out in the PR description.

9. **`WithCACertFile` reads at option-creation time** — If the CA cert is rotated on disk, the client won't pick it up. Worth documenting in the godoc.

### What looks good

- Clean functional options pattern for the new options
- Good CA cert merging logic in `WithCACertFile` (handles existing config, system pool fallback)
- `NewClientPoolWithOptions` alongside existing `NewClientPool` for backward compat
- Solid test coverage including edge cases (invalid PEM, empty file, etc.)
- TLS 1.2 minimum version enforcement

PR #872, opened 2026-01-28
Allow to override version id and modification count

merlimat:
During migrations to Oxia, it would be useful to preserve the original version id and modification counts of records.

This will only be exposed in Java client API.

### DIFF
--- a/oxiad/dataserver/database/db.go
+++ b/oxiad/dataserver/database/db.go
@@ -565,7 +565,11 @@ func (d *db) applyPut(batch kvstore.WriteBatch, baseVersionId *atomic.Int64, not
 				Status: status,
 			}, nil
 		}
-		versionId = baseVersionId.Add(1)
+		if putReq.OverrideVersionId != nil {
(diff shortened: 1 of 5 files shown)


oxiad/dataserver/database/db.go
```diff
@@ -565,7 +565,11 @@ func (d *db) applyPut(batch kvstore.WriteBatch, baseVersionId *atomic.Int64, not
 				Status: status,
 			}, nil
 		}
-		versionId = baseVersionId.Add(1)
+		if putReq.OverrideVersionId != nil {
```
mattisonchao, line 568: This will introduce the inconsistency of the version ID when rolling upgrade the Oxia DB. we need to find a way to avoid it.

merlimat: That's a good point. I don't think it should be a big problem though: 
 1. The feature is only meant for a 1-shot migration
 2. Oxia should have been already upgraded before attempting to use it

mattisonchao (approved):

PR #881, opened 2026-02-05
Fix type assertion panic in Cache.Get() when key not found

labuladong:
## Summary

Fix a type assertion panic that occurs when `Cache.Get()` is called for a key that doesn't exist in Oxia.

## Problem

When a key is not found, the cache stores a negative cache entry using:
```go
cr := empty[cachedResult[Value]]()
```

However, `cachedResult[Value]` is a type alias for `Optional[valueVersion[Value]]`, so this creates a double-wrapped optional type:
```
*optional[Optional[valueVersion[Value]]]
```

When `Cache.Get()` later retrieves this entry and attempts the type assertion:
```go
cachedValue.(cachedResult[Value])
```

It fails with:
```
interface conversion: *oxia.optional[cachedResult[...]] is not cachedResult[...]: missing method Get
```

## Solution

Use `empty[valueVersion[Value]]()` instead, which creates `*optional[valueVersion[Value]]` - the same type structure as the non-empty case, ensuring type consistency.

## Type Analysis

```
cachedResult[Value] = Optional[valueVersion[Value]]  // type alias

// Before (BUG):
empty[cachedResult[Value]]() 
= empty[Optional[valueVersion[Value]]]()
= *optional[Optional[valueVersion[Value]]]  // double-wrapped!

// After (FIX):
empty[valueVersion[Value]]()
= *optional[valueVersion[Value]]  // correct, matches non-empty case
```

### DIFF
--- a/oxiad/coordinator/controller/shard_controller_election.go
+++ b/oxiad/coordinator/controller/shard_controller_election.go
@@ -146,7 +146,7 @@ func (e *ShardElection) fenceNewTermQuorum(term int64, ensemble []model.Server, 
 				}, func() {
 					entryId, err := e.fenceNewTerm(fencingContext, term, pinedServer)
 					if err != nil {
-						e.Warn("Failed to fenceNewTerm data server", slog.Any("error", err), slog.Any("data-server", pinedServer))
+						e.Warn("FenceNewTerm failed", slog.Any("error", err), slog.Any("data-server", pinedServer))
 					} else {
 						e.Info("Processed fenceNewTerm response", slog.Any("data-server", pinedServer), slog.Any("entry-id", entryId))
 					}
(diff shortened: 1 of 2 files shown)


oxia/cache.go
```diff
@@ -267,7 +267,12 @@ func (c *cacheImpl[Value]) Get(ctx context.Context, key string) (value Value, ve
 func (c *cacheImpl[Value]) load(ctx context.Context, key string) (value Value, version Version, err error) {
 	_, data, existingVersion, err := c.client.Get(ctx, key)
 	if errors.Is(err, ErrKeyNotFound) {
-		cr := empty[cachedResult[Value]]()
+		// Use valueVersion[Value] instead of cachedResult[Value] to ensure
```
mattisonchao: Good catch!

IMO, we could remove this multi-line comment and add a method test to cover it.

mattisonchao (approved):

PR #894, opened 2026-02-23
fix: re-sort node load ratios after recalculation

mattisonchao:
### Motivation

After `MoveShardToNode` + `ReCalculateRatios`, the node ratios are updated but the list is **not re-sorted**. `lowerestLoadSelector` iterates from index 0 expecting ascending order, so it picks the first node (no longer the lowest) instead of the actual lowest-load node.

### Modification

- Add a sort call at the end of `Ratio.ReCalculateRatios()` in `load_ratio.go` to maintain ascending order after ratio updates
- Add reproduction test `TestSelectLowerestLoadSelectorAfterShardMove` that moves shards to make the first node no longer the lowest, then verifies the selector picks the actual lowest-load node

### DIFF
--- a/oxiad/coordinator/model/load_ratio.go
+++ b/oxiad/coordinator/model/load_ratio.go
@@ -81,26 +81,18 @@ func (r *Ratio) MoveShardToNode(shard *ShardLoadRatio, fromNode string, toNode s
 }
 
 func (r *Ratio) ReCalculateRatios() {
-	iter := r.nodeLoadRatios.Iterator()
-	if !iter.First() {
+	if r.nodeLoadRatios.Empty() {
 		return
-	}
-	nodeRatio := iter.Value()
-	r.maxNodeLoadRatio = nodeRatio.Ratio
-	r.minNodeLoadRatio = nodeRatio.Ratio
-
-	for iter.Next() {
-		nodeLoadRatio := iter.Value()
-		if nodeLoadRatio.Ratio > r.maxNodeLoadRatio {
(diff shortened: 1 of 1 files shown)


mattisonchao:
## Review Summary

### Problem

`balanceHighestNode()` and `cleanDeletedNode()` call `swapShard()` in a loop. Each `swapShard()` call:
1. Uses `lowerestLoadSelector.Select()` which returns the **first candidate** from the node list (assumes ascending sort order)
2. Calls `MoveShardToNode()` — mutates ratios in place, **does not re-sort**
3. Calls `ReCalculateRatios()` — updates min/max, **did not re-sort**

After the first iteration, the list order is stale, causing subsequent selector calls to pick the wrong node.

### Fix

Sort the node list in `ReCalculateRatios()` before reading min/max. This keeps the sorted invariant that `lowerestLoadSelector` and `balanceHighestNode` (which iterates from `Last()` for the highest node) both depend on.

### Changes

| File | Change |
|------|--------|
| `oxiad/coordinator/model/load_ratio.go` | Sort `nodeLoadRatios` in `ReCalculateRatios()`, then read first/last for min/max |
| `oxiad/coordinator/selector/single/lowerest_load_selector_test.go` | Add `TestSelectLowerestLoadSelectorAfterShardMove` reproducing the bug |

### Test Results

All coordinator tests pass with no regressions:

- `coordinator/action` ✅
- `coordinator/controller` ✅
- `coordinator/metadata` ✅
- `coordinator/model` ✅
- `coordinator/selector/ensemble` ✅
- `coordinator/selector/single` ✅
- `coordinator/util` ✅

PR #910, opened 2026-02-25
fix: improve leader rebalance candidate selection to prevent stuck states

mattisonchao:
## Summary

- Fixes flaky `TestLeaderBalancedNodeAdded` which failed ~10-40% of runs due to leader rebalancing getting permanently stuck
- Root cause: the "no valid candidates" check was too conservative — it quarantined shards when the best candidate had `maxLeaders-1` leaders, preventing useful leader shuffling that enables convergence
- Three changes in `rebalanceLeader()`:
  1. **Skip current leader in candidate evaluation** — only consider other ensemble members when computing `minCandidateLeaders`
  2. **Relax quarantine threshold** — changed from `minCandidateLeaders+1 == maxLeaders` to `minCandidateLeaders >= maxLeaders`, allowing moves to nodes with one fewer leader
  3. **Process multiple shards per trigger** — continue iterating after both election success (`maxLeaders--; continue`) and failure (`continue`) instead of breaking

## Test plan

- [x] `TestLeaderBalancedNodeAdded` — 63/63 consecutive passes with `-race` (previously ~10-40% failure rate)
- [x] `TestLeaderBalanced` — 3/3 passes
- [x] `TestLeaderBalancedNodeCrashAndBack` — 3/3 passes
- [x] `go test -race ./oxiad/coordinator/balancer/...` — all unit tests pass

### DIFF
--- a/oxiad/coordinator/balancer/scheduler.go
+++ b/oxiad/coordinator/balancer/scheduler.go
@@ -417,6 +417,9 @@ func (r *nodeBasedBalancer) rebalanceLeader() {
 		minCandidateLeaders := -1
 		for _, candidate := range shardStatus.Ensemble {
 			candidateID := candidate.GetIdentifier()
+			if candidateID == maxLeadersNodeID {
+				continue
+			}
 			if !r.nodeAvailableJudger(candidateID) {
 				continue
 			}
@@ -426,7 +429,7 @@ func (r *nodeBasedBalancer) rebalanceLeader() {
 				minCandidateLeaders = leaders
 			}
 		}
-		if minCandidateLeaders == maxLeaders || minCandidateLeaders+1 == maxLeaders {
+		if minCandidateLeaders == -1 || minCandidateLeaders >= maxLeaders {
 			r.Info("quarantine the shard due to no valid candidates", slog.Int64("shard", shard), slog.Any("leader", maxLeadersNodeID))
 			r.shardQuarantineShardMap.Store(shard, time.Now())
 			continue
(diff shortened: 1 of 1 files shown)


coderzc:
## Code Review

### Overview
This PR fixes a flaky test (`TestLeaderBalancedNodeAdded`) by improving the leader rebalancing logic in the Oxia coordinator. The root cause was overly conservative candidate selection that could permanently quarantine shards, preventing convergence.

---

### Changes Analysis

**File:** `oxiad/coordinator/balancer/scheduler.go`

#### 1. Skip current leader in candidate evaluation (lines 417-420)
```go
if candidateID == maxLeadersNodeID {
    continue
}
```
**Good:** Excludes the current leader when computing `minCandidateLeaders`, ensuring we only consider actual move candidates. This makes the min calculation more accurate.

#### 2. Relax quarantine threshold (line 429)
```diff
- if minCandidateLeaders == maxLeaders || minCandidateLeaders+1 == maxLeaders {
+ if minCandidateLeaders == -1 || minCandidateLeaders >= maxLeaders {
```
**Good:** The new logic is cleaner and correctly handles the case where no valid candidates exist (`-1`). The `>=` check is equivalent to the previous two conditions but more readable.

#### 3. Process multiple shards per trigger (lines 447-453)
```go
continue  // instead of quarantine + break
// ...
maxLeaders--
continue  // instead of break
```
**Good:** Allows processing multiple shards in a single rebalancing pass, improving convergence speed.

---

### Review Feedback

#### ✅ Positive Aspects
- **Clear problem diagnosis** — PR description explains the root cause well
- **Minimal, targeted fix** — Only 7 lines added, 3 removed
- **Good test coverage** — 63 consecutive passes with race detector
- **Improved readability** — The `>=` condition is cleaner than the original

#### ⚠️ Suggestions

1. **Log message accuracy** — The log on line 430 says "no valid candidates" but the condition `minCandidateLeaders >= maxLeaders` means there ARE candidates, just none that improve balance. Consider updating the message to reflect this.

2. **Edge case consideration** — When `minCandidateLeaders == maxLeaders`, moving to an equally loaded node could be useful for ensemble health (e.g., if current node is going down). Is this intentionally prevented?

3. **Comment update** — The comment on line 449 says "no changes" but the behavior changed: previously it quarantined, now it continues. Update the comment to reflect the new intent.

#### ✅ Verification
- [x] Race detector passes
- [x] Related tests pass
- [x] Logic appears correct

---

### Summary
**Approve with minor suggestions.** The fix is well-targeted, thoroughly tested, and improves both correctness and readability. Consider updating the log message and comment for clarity.

PR #933, opened 2026-03-08
fix: only call GetInfo RPC on state transition, not every health check

### LINKED ISSUE #932: Redundant GetInfo RPC called on every health check instead of only on state transitions
## Summary

The coordinator's `dataServerController` sends a `GetInfo` RPC to every data server on **every successful health check** (~every 2 seconds), instead of only when the server transitions from `NotRunning` to `Running`. This generates excessive log noise and unnecessary network traffic.

## Root Cause

In `oxiad/coordinator/controller/dataserver_controller.go`:

1. `healthCheckHandler()` (line 328) calls `becomeAvailable()` on every successful health check response
2. `becomeAvailable()` (line 284) unconditionally calls `syncDataServerInfo()` after setting the status to `Running` — even when the status was **already** `Running`
3. `syncDataServerInfo()` (line 311) issues a `GetInfo` RPC every time it's called

The state-transition-specific logic (clearing pooled connections, resetting backoffs) is correctly gated behind `if n.status == NotRunning`, but the `syncDataServerInfo()` call is outside that gate.

## Impact

- Every data server receives a `GetInfo` RPC every ~2 seconds from the coordinator, continuously
- Logs are flooded with `Received GetInfo request` messages at INFO level
- Unnecessary network overhead on the internal RPC path

## Observed Behavior

```
Mar  8 14:25:18 INF Received GetInfo request component=internal-rpc-server peer=10.244.0.20:60248 request={}
Mar  8 14:25:20 INF Received GetInfo request component=internal-rpc-server peer=10.244.0.20:60248 request={}
Mar  8 14:25:22 INF Received GetInfo request component=internal-rpc-server peer=10.244.0.20:60248 request={}
... (repeats every ~2 seconds indefinitely)
```

## Suggested Fix

In `becomeAvailable()`, only call `syncDataServerInfo()` when there is an actual state transition from `NotRunning` → `Running`. When the server is already `Running`, skip the redundant `GetInfo` call.

```go
func (n *dataServerController) becomeAvailable() {
	n.statusLock.Lock()
+	wasNotRunning := n.status == NotRunning
	if n.status == NotRunning {
		n.Info("Storage data server is back online")
		n.rpc.ClearPooledConnections(n.dataServer)
		n.healthCheckBackoff.Reset()
		n.healthWatchBackoff.Reset()
	}
	n.status = Running
	n.statusLock.Unlock()

+	if !wasNotRunning {
+		return
+	}
+
	// sync the latest info
	bo := commontime.NewBackOffWithInitialInterval(n.ctx, defaultInitialRetryBackoff)
	_ = backoff.RetryNotify(func() error {
		return n.syncDataServerInfo()
	}, bo, func(err error, duration time.Duration) {
		...
	})
}
```

mattisonchao:
## Motivation

The coordinator's `dataServerController` sends a `GetInfo` RPC to every data server on **every successful health check** (~every 2s), flooding logs with `Received GetInfo request` and generating unnecessary network traffic. `GetInfo` only needs to be called when a server transitions from `NotRunning` to `Running`.

## Modifications

- **`becomeAvailable()` → early return**: When status is already `Running`, skip `syncDataServerInfo()`. Only call `GetInfo` on actual `NotRunning` → `Running` transitions.
- **Start controller as `NotRunning`**: The first health check naturally triggers the transition and initial `GetInfo` call — no extra init goroutine needed.
- **`ConcurrentBackOff`**: Mutex-wrapped `backoff.BackOff` shared by both `healthPingWithRetries` and `healthWatchWithRetries`. Fixes a latent data race where `becomeAvailable()` reset backoff objects concurrently with `backoff.RetryNotify`.
- **Immediate health ping**: Do an immediate health check on startup instead of waiting for the 2s ticker, eliminating unnecessary startup delay.
- **Increased test timeouts**: `TestLeaderHint*` timeouts increased from 10s to 20s for CI stability.

Closes #932

## Test plan
- [x] All tests pass with `-race`
- [x] New `TestDataServerController_GetInfoOnlyCalledOnStateTransition` verifies `GetInfo` only fires on `NotRunning` → `Running` transitions and stays stable while already `Running`

### DIFF
--- a/oxiad/coordinator/controller/dataserver_controller.go
+++ b/oxiad/coordinator/controller/dataserver_controller.go
@@ -202,10 +202,21 @@ func (n *dataServerController) sendAssignmentsDispatchWithRetries() {
 	})
 }
 
+func (n *dataServerController) doHealthPing() error {
+	pingCtx, pingCancel := context.WithTimeout(n.ctx, healthCheckProbeTimeout)
+	response, err := n.healthClient.Check(pingCtx, &grpc_health_v1.HealthCheckRequest{Service: ""})
+	pingCancel()
+	return n.healthCheckHandler(response, err)
+}
+
 func (n *dataServerController) healthPingWithRetries() {
 	defer n.Done()
 	_ = backoff.RetryNotify(func() error {
 		n.maybeInitHealthClient()
+		// Immediate check on startup instead of waiting for first tick
+		if err := n.doHealthPing(); err != nil {
+			return err
+		}
 		ticker := time.NewTicker(healthCheckProbeInterval)
 		defer ticker.Stop()
 		for {
@@ -213,10 +224,7 @@ func (n *dataServerController) healthPingWithRetries() {
 			case <-n.ctx.Done():
 				return nil
 			case <-ticker.C:
-				pingCtx, pingCancel := context.WithTimeout(n.ctx, healthCheckProbeTimeout)
-				response, err := n.healthClient.Check(pingCtx, &grpc_health_v1.HealthCheckRequest{Service: ""})
-				pingCancel()
-				if err := n.healthCheckHandler(response, err); err != nil {
+				if err := n.doHealthPing(); err != nil {
 					n.Warn("Data server stopped responding to ping")
 					return err
 				}


mattisonchao (commented):
**Self-review — ready for review.**

**Summary of changes:**

1. **Fix redundant GetInfo RPCs**: `becomeAvailable()` now early-returns when status is already `Running`, so `syncDataServerInfo()` only runs on `NotRunning` → `Running` transitions. This eliminates the ~2s polling of GetInfo on every health check.

2. **Start with `NotRunning`**: Controller initial status changed from `Running` to `NotRunning`, so the first health check naturally triggers the transition and initial `GetInfo` call — no extra init goroutine needed.

3. **Fix data race on backoff**: Introduced `ConcurrentBackOff` (mutex-wrapped `backoff.BackOff`) shared by both `healthPingWithRetries` and `healthWatchWithRetries`. This fixes a latent race where `becomeAvailable()` reset backoff objects concurrently with `backoff.RetryNotify` — previously hidden because the controller started as `Running` and never entered the reset path on first health check.

4. **Test**: `TestDataServerController_GetInfoOnlyCalledOnStateTransition` verifies GetInfo count stays stable while Running and only increases on state transitions.

All tests pass with `-race`.

PR #945, opened 2026-03-11
feat: add data server support for shard split observers and snapshots

merlimat:
Implement observer follower cursors that replicate data from a parent shard to child shards during split operations:
- Observer cursors using the parent's term for proper term validation on the receiving child (coordinator fences child with parent's term)
- SendSnapshot/Replicate flow for shard-to-shard data bootstrap
- Hash range propagation via gRPC metadata so child followers activate receiver-side filtering (SetSplitHashRange)
- FilterDBForSplit and ApplyLogEntryWithSplitFilter for key filtering based on hash range boundaries
- RemoveObserver RPC: cleanly remove observer cursors from parent leader without fencing, used during split abort

Extend quorum ack tracker to support adding/removing followers dynamically during split lifecycle. Add follower controller support for receiving observer snapshots and replication streams.

### DIFF
--- a/oxiad/dataserver/controller/follow/follower_controller.go
+++ b/oxiad/dataserver/controller/follow/follower_controller.go
@@ -554,8 +554,6 @@ func (fc *followerController) InstallSnapshot(stream proto.OxiaLogReplication_Se
 	case term != constant.I64NegativeOne && firstChunk.Term != constant.I64NegativeOne && term != firstChunk.Term:
 		// The follower could be left with term=-1 by a previous failed
 		// attempt at sending the snapshot. It's ok to proceed in that case.
-		// Observer cursors (used during shard splits) send term=-1 to bypass
-		// term validation since the child shard has its own term.
 		err = dserror.ErrInvalidTerm
 		return err
 	}
@@ -610,8 +608,11 @@ func (fc *followerController) InstallSnapshot(stream proto.OxiaLogReplication_Se
 	// keys within the child's hash range.
 	if fc.splitHashRange != nil {
 		if err = database.FilterDBForSplit(db.RawKV(), *fc.splitHashRange); err != nil {
-			return errors.Wrap(err, "failed to filter snapshot for split")
+			return errors.Wrapf(multierr.Combine(dserror.ErrResourceNotAvailable, err), "failed to filter snapshot for split")
 		}
+		// FilterDBForSplit deletes the checksum key (it's invalid after
+		// filtering), so reset the in-memory cached checksum state.
+		db.ResetChecksum()
 	}
 
 	if err = stream.SendAndClose(&proto.SnapshotResponse{
--- a/oxiad/dataserver/controller/lead/leader_controller.go
+++ b/oxiad/dataserver/controller/lead/leader_controller.go
@@ -534,7 +534,7 @@ func (lc *leaderController) RemoveObserver(req *proto.RemoveObserverRequest) (*p
 
 func (lc *leaderController) addFollower(follower string, followerHeadEntryId *proto.EntryId) error {
 	term := lc.term.Load()
-	followerHeadEntryId, err := lc.truncateFollowerIfNeeded(follower, followerHeadEntryId)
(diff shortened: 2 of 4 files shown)


oxiad/dataserver/controller/follow/follower_controller.go
```diff
@@ -581,6 +606,14 @@ func (fc *followerController) InstallSnapshot(stream proto.OxiaLogReplication_Se
 	fc.lastAppendedOffset.Store(rawCommitOffset)
 	fc.advertisedCommitOffset.Store(rawCommitOffset)
 
+	// If this follower is a split child, filter the snapshot to only retain
+	// keys within the child's hash range.
+	if fc.splitHashRange != nil {
+		if err = database.FilterDBForSplit(db.RawKV(), *fc.splitHashRange); err != nil {
+			return errors.Wrap(err, "failed to filter snapshot for split")
```
mattisonchao: Suggest giving a `dserror.ErrResourceNotAvailable` for a clearer error abstraction.

oxiad/dataserver/internal_rpc_server.go
```diff
@@ -326,6 +354,11 @@ func (s *internalRpcServer) Replicate(srv proto.OxiaLogReplication_ReplicateServ
 		return err
 	}
 
+	// Activate split filtering if hash range metadata is present
```
mattisonchao, line 357: I am considering whether we could make the truncate become `prepare` for the follower to help us initialise some states to avoid write same logic in both `snapshot` and `replication.`

mattisonchao (dismissed):

oxiad/dataserver/controller/lead/follower_cursor.go
```diff
@@ -161,6 +163,82 @@ func NewFollowerCursor( //nolint:revive
...
+	fc.ctx, fc.cancel = context.WithCancel(context.Background())
+	fc.backoff = time2.NewBackOff(fc.ctx)
+
+	fc.lastPushed.Store(ackOffset)
+	fc.ackOffset.Store(ackOffset)
+
+	// Observer uses a no-op cursor acker — its acks don't affect quorum
+	fc.cursorAcker = NewNoOpCursorAcker()
+
+	go process.DoWithLabels(
```
mattisonchao, line 227: We need to make the goroutine lifecycle controlled by the follower_cursor. but it seems like we don't have this kind of logic now.  let's do it in other PRs.

oxiad/dataserver/controller/lead/leader_controller.go
```diff
@@ -512,6 +577,51 @@ func (lc *leaderController) addFollower(follower string, followerHeadEntryId *pr
 	return nil
 }
 
+func (lc *leaderController) addObserver(observerKey string, follower string, followerHeadEntryId *proto.EntryId,
+	targetShardId int64, splitHashRange *proto.Int32HashRange) error {
+	followerHeadEntryId, err := lc.truncateFollowerIfNeeded(follower, followerHeadEntryId)
```
mattisonchao: It seems harmless right now, but it's better to pass the `targetShard` to truncate.

oxiad/dataserver/controller/follow/follower_controller.go
```diff
@@ -528,9 +551,11 @@ func (fc *followerController) InstallSnapshot(stream proto.OxiaLogReplication_Se
 		return errors.Wrapf(multierr.Combine(dserror.ErrResourceNotAvailable, err), "failed to read first snapshot chunk")
 	case firstChunk == nil:
 		return nil
-	case term != constant.I64NegativeOne && term != firstChunk.Term:
+	case term != constant.I64NegativeOne && firstChunk.Term != constant.I64NegativeOne && term != firstChunk.Term:
 		// The follower could be left with term=-1 by a previous failed
 		// attempt at sending the snapshot. It's ok to proceed in that case.
+		// Observer cursors (used during shard splits) send term=-1 to bypass
```
mattisonchao: I didn't see `observer_cursor` set the `term=-1`.

PR #978, opened 2026-03-29
feat: add log format option to server and coordinator config

mattisonchao:
### Motivation

Server and coordinator need a way to configure log format (text/json) via config file, with proper precedence over CLI flags.

### Modification

- Add `Format` field to `LogOptions` with `LogFormatText` and `LogFormatJSON` constants
- Apply config file log format at startup for server and coordinator (config file overrides CLI `--log-json`, absent means CLI preserved)
- Add `format` field to YAML configs and JSON schemas for both dataserver and coordinator
- Remove unused `DefaultLogLevel` from option package (CLI default in `logging` package is the source of truth)
- Keep `LogOptions.WithDefault()` empty — level and format use "empty means not configured" semantics to avoid overriding CLI values

### DIFF
--- a/oxiad/common/option/option.go
+++ b/oxiad/common/option/option.go
@@ -20,10 +20,7 @@ import (
 	"go.uber.org/multierr"
 )
 
-const (
-	DefaultMetricsPort = 8080
-	DefaultLogLevel    = "info"
-)
+const DefaultMetricsPort = 8080
 
 type ObservabilityOptions struct {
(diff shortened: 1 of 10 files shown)


oxiad/common/option/option.go
```diff
@@ -64,14 +61,18 @@ func (*MetricOptions) Validate() error {
 	return nil
 }
 
+const (
+	LogFormatText    = "text"
```
merlimat, line 65: technically `json` is also text. can we use `plain`

PR #1021, opened 2026-04-11
fix: publish namespace status during initial placement

stanislavkozlovski:
During a load test of Oxia, I ran into some rebalance storms. The proper workaround for me was to warm up easier -  create the shards, then wait before starting to blast the system with operations.

I was confused why rebalances were happening when I'm simply creating a few shards from scratch. This led me to this bug, where the `ensembleSupplier` would not get an ever-updating clusterStatus (which it expects to pass to `GroupingShardsNodeByStatus` in the one prod code path). By passing an empty clusterStatus as it did, the initial load balancing placement could never work.

### DIFF
--- a/oxiad/coordinator/util/cluster_updates_test.go
+++ b/oxiad/coordinator/util/cluster_updates_test.go
@@ -25,6 +25,79 @@ import (
 	"github.com/oxia-db/oxia/oxiad/coordinator/model"
 )
 
+// loadAwareEnsembleSupplier picks the `nc.ReplicationFactor` servers that
+// currently hold the fewest ensemble slots across all already-placed shards in
+// `cs`. It is used to verify that ApplyClusterChanges publishes intermediate
+// shard placements into status during the init loop, so that successive
+// per-shard ensemble selections within the same init cycle observe each
+// other's choices. Without that publication every call sees an empty cluster
+// and returns the same servers, producing wildly uneven initial placement.
+func loadAwareEnsembleSupplier(servers []model.Server) func(*model.NamespaceConfig, *model.ClusterStatus) ([]model.Server, error) {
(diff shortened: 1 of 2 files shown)


oxiad/coordinator/util/cluster_updates_test.go
```diff
@@ -25,6 +25,80 @@ import (
...
+
+// TestClientUpdates_InitialPlacementSeesPriorShards verifies that the
+// per-shard ensembleSupplier called by ApplyClusterChanges during a new
+// namespace's bootstrap can observe shards that were placed earlier in the
+// same call. Regression for an upstream bug where the new NamespaceStatus was
+// only assigned into status.Namespaces *after* the GenerateShards loop, so
+// every shard was selected against an empty cluster snapshot, leading to
+// pseudo-random initial placement that immediately tripped the count-based
+// load-balancer once the cluster came up.
+func TestClientUpdates_InitialPlacementSeesPriorShards(t *testing.T) {
```
stanislavkozlovski, line 65: tested and confirm the test fails on the `main` branch, and pass with the change

oxiad/coordinator/util/cluster_updates_test.go
```diff
@@ -25,6 +25,79 @@ import (
...
+		slots[s.Internal] = 0
+	}
+	for _, sh := range status.Namespaces["ns-1"].Shards {
+		assert.Len(t, sh.Ensemble, rf)
+		for _, m := range sh.Ensemble {
+			slots[m.Internal]++
+		}
+	}
+
+	// shardCount * rf == 48 ensemble slots distributed across 4 servers ⇒
```
stanislavkozlovski, line 91: Without the fix, every call would return the same first `rf` (3) servers, so some would receive 16 slots and the others 0.

mattisonchao (approved):
LGTM

PR #1029, opened 2026-04-12
security: add optional auth config to internal and admin servers

merlimat:
## Summary
- Add `Auth` field to `InternalServerOptions` in both dataserver and coordinator option packages, following the same pattern used by `PublicServerOptions`
- Add `Auth` field to `AdminServerOptions` in the coordinator option package
- Wire the new auth options through to `newInternalRpcServer()` and the coordinator's gRPC server constructors, replacing the hardcoded `&auth.Disabled`
- Update `WithDefault()` and `Validate()` methods for all modified option structs

## Test plan
- [x] `go build ./oxiad/... ./common/...` passes
- [x] `golangci-lint run ./oxiad/dataserver/... ./oxiad/coordinator/...` passes (no new issues)
- [x] Existing `TestInternalHealthCheck` updated and passes
- [ ] Verify that omitting the `auth` config in YAML still defaults to disabled (backward compatible)
- [ ] Verify that setting `auth` config on internal/admin servers enables authentication

### DIFF
--- a/oxiad/common/metric/metrics.go
+++ b/oxiad/common/metric/metrics.go
@@ -16,6 +16,7 @@ package metric
 
 import (
 	"context"
+	libtls "crypto/tls"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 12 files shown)


mattisonchao (approved):

mattisonchao:
hmm.. 

we need to support configuring the coordinator internally client auth/TLS in the following.

PR #1033, opened 2026-04-12
security: add input validation and DoS prevention for gRPC

merlimat:
## Summary
- Reduce max gRPC frame size from 256 MB to 16 MB to limit memory consumption from oversized messages
- Add `grpc.MaxConcurrentStreams(1000)` to prevent stream-based resource exhaustion
- Add write request validation enforcing max key size (10 KB), value size (1 MB), and batch counts (10k puts, 10k deletes, 100 delete ranges)
- Wire validation into both the unary `Write` RPC and the `WriteStream` path, before reaching the leader controller
- Add context cancellation check at the top of the write stream loop for prompt exit on cancelled streams

## Test plan
- [x] `go build ./oxiad/... ./common/...` passes
- [x] `golangci-lint run ./oxiad/dataserver/... ./oxiad/common/rpc/...` passes (no new issues)
- [x] `go test -v -run TestValidateWriteRequest ./oxiad/dataserver/...` -- all 5 tests pass
- [ ] Verify existing integration tests still pass with the reduced 16 MB frame size
- [ ] Verify that oversized keys/values are rejected with `InvalidArgument` status

### DIFF
--- a/oxiad/common/rpc/grpc_server.go
+++ b/oxiad/common/rpc/grpc_server.go
@@ -35,10 +35,11 @@ import (
 )
 
 const (
-	maxGrpcFrameSize                         = 16 * 1024 * 1024
-	defaultGrpcServerKeepAliveMinTime        = 5 * time.Second
-	defaultGrpcServerKeepPermitWithoutStream = true
-	ReadinessProbeService                    = "oxia-readiness"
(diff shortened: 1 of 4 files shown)


mattisonchao:
Hi @merlimat 

The SDK side also has similar configurations for frame size.  We should make it configurable to avoid some clusters that upgrade can't get data from the server.

PR #1037, opened 2026-04-13
Add get-data-server admin command

mattisonchao:
## Summary
- add a `GetDataServer` admin RPC and wire it through the admin client
- add the `oxia admin get-data-server` CLI command and its mock/test coverage
- add coordinator-side tests for found and not-found lookups

## Testing
- `go test ./cmd/admin/getdataserver ./oxia ./oxiad/coordinator`
- `make test` was started and progressed through repo-wide package runs, but the final `go test -cover -race ...` process stalled without reporting a failure in this environment

### DIFF
--- a/common/proto/admin.proto
+++ b/common/proto/admin.proto
@@ -25,6 +25,7 @@ option java_multiple_files = true;
 service OxiaAdmin {
 
   rpc ListDataServers(ListDataServersRequest) returns (ListDataServersResponse);
+  rpc GetDataServer(GetDataServerRequest) returns (DataServer);
 
   rpc ListNamespaces(ListNamespacesRequest) returns (ListNamespacesResponse);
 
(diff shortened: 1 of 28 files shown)


common/proto/admin.proto
```diff
@@ -46,12 +49,18 @@ message DataServer {
...
+  map<string, string> metadata = 4;
+  repeated replication.Feature supported_features = 5;
 }
 
 message ListDataServersResponse {
   repeated DataServer data_servers = 1;
 }
 
+message GetDataServerRequest {
+  string data_server = 1;
```
merlimat, line 61: is this supposed to be the `name`?  what if name is not set?

mattisonchao: primary name,  fallback to the internal address.

mattisonchao: Need more round iteration for this PR.

PR #1047, opened 2026-04-15
[fix] fix raft metadata

dao-jun:
### Motivation

  The coordinator's Raft metadata provider had several critical issues:

  1. Data races: stateContainer.Get() returned direct references to internal state, allowing callers to bypass locks and mutate FSM memory, causing state corruption under concurrent access.
  2. Split-brain risk: When a node lost Raft leadership, the coordinator continued running and could simultaneously perform shard elections and state updates alongside the new leader.
  3. Silent failures: StatusResource methods (Update, DeleteShardMetadata, etc.) silently discarded persistence errors. backoff.RetryNotify results were dropped (_ =), causing infinite retries when a node lost leadership with no way to terminate.
  4. Resource leaks: NewMetadataProviderRaft did not clean up already-created transport/store on initialization failure. Close() did not close the transport.
  5. Data corruption: DeleteRange caused integer overflow when maxInclusive == MaxUint64, producing an incorrect delete range. Snapshot.Persist did not check for short writes and called Cancel() on the success path.

### Changes

  Thread Safety & Data Isolation

  - Changed stateContainer from sync.Mutex to sync.RWMutex; added CloneState() returning a deep copy.
  - Get() now returns a Clone() copy to prevent external mutation of internal state.
  - ShardMetadata.Clone() performs deep copies of Leader, Ensemble, and RemovedNodes (including Name pointers).
  - Added Server.Clone() method.

  Leader Verification & Consistency

  - Get() and Store() now call raft.VerifyLeader() before operating to ensure the node is still the leader.
  - Get() calls raft.Barrier() to guarantee linearizable reads.
  - Improved WaitToBecomeLeader() to wait for leader state confirmation, then wait for all committed logs to be applied.

  Leadership Loss Detection

  - Added LeadershipAwareProvider interface exposing a LeadershipLost() channel.
  - Watches Raft NotifyCh for leadership transitions and signals the coordinator on loss.
  - coordinator.monitorLeadershipLoss() calls cancel() on loss, terminating all control loops.

  Error Propagation

  - StatusResource methods now return error: Update, UpdateShardMetadata, DeleteShardMetadata, ApplyChanges.
  - All backoff.RetryNotify calls are bound to the coordinator's context, stopping retries on cancellation.
  - Callers (shard_controller, split_controller, election) now check and handle errors appropriately.

  Resource Management

  - NewMetadataProviderRaft uses a defer cleanup to release raft node/transport/store on initialization failure.
  - Close() uses sync.Once and now also closes the transport.
  - kvRaftStore closes the factory if NewKV fails.
  - StoreLogs uses defer to close the WriteBatch.

  Snapshot Persistence Fixes

  - Checks sink.Write return length to detect short writes.
  - Success path calls only sink.Close(); error path calls only sink.Cancel().
  - Restore properly merges decode and close errors.

  Boundary Condition Fixes

  - DeleteRange uses logKeyEnd when maxInclusive == MaxUint64 to avoid integer overflow.
  - GetLog returns raft.ErrLogNotFound when the key does not exist.

  Configuration Validation

  - Validates raft-address as host:port format.
  - Requires raft-data-dir and raft-bootstrap-nodes to be set.
  - Checks bootstrap nodes for duplicates and ensures the local address is included.

  Tests

  - 3-node Raft cluster restart tests (metadata_raft_test.go).
  - FSM snapshot/restore, short write, and close error tests (metadata_raft_fsm_test.go).
  - Store GetLog missing, DeleteRange overflow, and factory cleanup tests (metadata_raft_store_test.go).
  - Configuration validation tests (option_test.go).
  - Context cancellation stops retry test (status_resource_test.go).
  - Coordinator startup leadership loss test (coordinator_leadership_test.go).
  - Multi-node coordinator restart end-to-end test (coordinator_raft_metadata_e2e_test.go).

### DIFF
--- a/oxiad/coordinator/balancer/scheduler_test.go
+++ b/oxiad/coordinator/balancer/scheduler_test.go
@@ -39,17 +39,22 @@ func (m *mockStatusResource) Load() *model.ClusterStatus {
 	return m.status
 }
 
-func (m *mockStatusResource) ApplyChanges(_ *model.ClusterConfig, _ resource.EnsembleSupplier) (*model.ClusterStatus, map[int64]string, []int64) {
-	return m.status, nil, nil
+func (m *mockStatusResource) ApplyChanges(_ *model.ClusterConfig, _ resource.EnsembleSupplier) (*model.ClusterStatus, map[int64]string, []int64, error) {
+	return m.status, nil, nil, nil
 }
 
-func (m *mockStatusResource) Update(newStatus *model.ClusterStatus) {
+func (m *mockStatusResource) Update(newStatus *model.ClusterStatus) error {
 	m.status = newStatus
+	return nil
 }
 
-func (m *mockStatusResource) UpdateShardMetadata(_ string, _ int64, _ model.ShardMetadata) {}
+func (*mockStatusResource) UpdateShardMetadata(_ string, _ int64, _ model.ShardMetadata) error {
+	return nil
+}
 
-func (m *mockStatusResource) DeleteShardMetadata(_ string, _ int64) {}
+func (*mockStatusResource) DeleteShardMetadata(_ string, _ int64) error {
+	return nil
+}
 
 func (m *mockStatusResource) IsReady(_ *model.ClusterConfig) bool {
 	return true
--- a/oxiad/coordinator/controller/shard_controller.go
+++ b/oxiad/coordinator/controller/shard_controller.go
@@ -427,7 +427,9 @@ func (s *shardController) deleteShard() error {
 		)
 	}
 
-	s.statusResource.DeleteShardMetadata(s.namespace, s.shard)
+	if err := s.statusResource.DeleteShardMetadata(s.namespace, s.shard); err != nil {
+		return err
+	}
 	s.eventListener.ShardDeleted(s.shard)
 	return s.close()
 }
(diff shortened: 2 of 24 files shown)


mattisonchao:
Hi @dao-jun 

Could you split the PR into multiple patches?

merlimat:
Hi @dao-jun — excellent analysis in this PR; while working through a coordinator correctness review we cross-checked every issue raised here against current `main` (after the coordinator `runtime`/`metadata/provider` refactor, this branch is unfortunately unrebaseable). Status of each issue group, and the plan:

**Already absorbed by the refactor since this PR was opened:**
- *FSM data races / deep clones*: the FSM now stores raw JSON documents and clones bytes on apply/snapshot; the cluster state moved to protobuf messages, so the hand-written `Clone()` concerns no longer apply.
- *Error propagation / unbounded retries*: the reconciler/watch architecture binds all backoff retries to the coordinator context.
- *`VerifyLeader` on `Store`* and *`Close` idempotency* are present at tip.

**Still valid — landing as the focused patches that were requested here, with credit to this PR:**
- #1170 — snapshot `Persist` calling `Cancel()` on the success path (snapshots never retained; metadata lost on restart after compaction).
- #1172 — resource lifecycle: init-failure cleanup in `New()`, transport close in `Close()`, write-batch close on `StoreLogs` error paths, KV-factory close, and the `DeleteRange` `MaxUint64` overflow.
- Next up: leadership lifecycle — `WaitToBecomeLeader` currently returns on *any* `LeaderCh` signal (including losses), there is no barrier for read linearizability after taking over, and nothing monitors leadership *loss*, so a deposed coordinator keeps its control loops running (your split-brain finding). The refactor's context plumbing now makes the cancellation side much simpler than it was when this PR was written.

Once that lands, every still-applicable fix from this PR will be on main, and this one can be closed. Thanks again — the issue inventory here was spot on.

dao-jun:
@merlimat Thanks for your work! I'll close the PR

PR #1076, opened 2026-04-29
fix: render `option.Duration` as human-readable string in JSON

merlimat:
## Summary
`option.Duration` already marshals to YAML as `"1h0m0s"` but defaulted to its underlying `int64` nanosecond representation when marshaled to JSON. This made JSON logs hard to read whenever a config containing a duration was logged.

Adding `MarshalJSON` makes the JSON form match the YAML form (and the JSONSchema, which already declares the type as a string).

Before:
```
"storage":{"wal":{...,"retention":3600000000000},...,"notification":{"retention":3600000000000}}
```

After:
```
"storage":{"wal":{...,"retention":"1h0m0s"},...,"notification":{"retention":"1h0m0s"}}
```

## Test plan
- [x] `make build` passes
- [x] `bin/oxia standalone -j` shows `"retention":"1h0m0s"` in the startup config log


mattisonchao (approved):

mattisonchao:
Hi @merlimat  

will you apply Copilot comments?

PR #1107, opened 2026-05-05
fix: keep raft snapshots rollback compatible

mattisonchao:
## Motivation
- Preserve rollback compatibility for Raft metadata snapshots after the metadata provider moved to a multi-document snapshot envelope.
- Keep v0.16.3 readers able to recover the cluster status from the legacy top-level snapshot fields.

## Modifications
- Include legacy `state` and `current_version` fields in Raft snapshot payloads alongside the current `documents` map.
- Populate the legacy fields from the cluster status document only.
- Added coverage that verifies snapshots contain both the current document map and the legacy status snapshot shape.

## Testing
- `go test ./oxiad/coordinator/metadata/provider/raft`
- `go test ./common/proto ./oxiad/coordinator/metadata/...`
- `make lint`
- `make license-check`
- `git diff --check`

### DIFF
--- a/oxiad/coordinator/metadata/provider/raft/raft_fsm.go
+++ b/oxiad/coordinator/metadata/provider/raft/raft_fsm.go
@@ -143,9 +143,17 @@ type documentContainer struct {
 }
 
 func marshalStateContainer(sc *stateContainer) ([]byte, error) {
-	return json.Marshal(persistedStateContainer{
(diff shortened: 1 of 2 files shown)


mattisonchao:
Closing this follow-up because we decided not to preserve legacy Raft snapshot rollback compatibility; the current multi-document snapshot format on main is the intended format.

PR #1156, opened 2026-06-03
feat: encode coordinator info in metadata lease

mattisonchao:
## Motivation
- Coordinator management API redirects need the elected coordinator's public management address.
- Kubernetes leader election currently stores only an opaque identity string in the lease holder.

## Modifications
- Add `server.public.advertisedAddress` and default it from the public bind address when unset.
- Add metadata protobuf `CoordinatorInfo` with coordinator identity and public address.
- Encode `CoordinatorInfo` into Kubernetes lease holder identity and provide decode support for future redirect logic.
- Wire configmap metadata providers to use coordinator identity plus advertised public address.
- Add tests for advertised address defaulting and coordinator lease info encoding.

## Testing
- `jq empty conf/schema/coordinator.json`
- `go test ./common/proto ./cmd/coordinator ./oxiad/coordinator/... ./tests/coordinator`


oxiad/coordinator/option/option.go
```diff
@@ -107,19 +107,34 @@ func (iso *InternalServerOptions) Validate() error {
...
 }
 
 func (pso *PublicServerOptions) WithDefault() {
+	defaultBindAddress := fmt.Sprintf("0.0.0.0:%d", constant.DefaultAdminPort)
 	if pso.BindAddress == "" {
-		pso.BindAddress = fmt.Sprintf("0.0.0.0:%d", constant.DefaultAdminPort)
+		pso.BindAddress = defaultBindAddress
+	}
+	if pso.AdvertisedAddress == "" {
+		pso.AdvertisedAddress = pso.BindAddress
```
mattisonchao, line 123: Can we avoid defaulting this directly to `bindAddress` here? `bindAddress` is commonly `0.0.0.0:6651`, which is valid for listening but not a reachable address for other coordinators. Since this value gets encoded into the coordinator lease and used for redirects, a safer optional default would be to derive `advertisedAddress` from `os.Hostname()` plus the public bind port when the user does not configure it explicitly.

PR #1166, opened 2026-06-11
fix: register the gRPC codec through the CodecV2 interface

merlimat:
### Motivation

The vtproto codec is registered as a legacy `encoding.Codec` (V1). Since grpc-go 1.66, V1 codecs get wrapped in a bridge whose `Unmarshal` calls `BufferSlice.Materialize()` — an unconditional allocate-and-copy of every received message — while the codec's `MarshalVT` allocates a fresh, unpooled buffer for every outgoing message. Every message of every RPC in the system (client writes/reads, replication appends/acks, assignments) pays one avoidable allocation and one full-payload copy in each direction. At replication rates this is hundreds of MB/s of garbage cluster-wide.

### Changes

- **grpc-go upgraded v1.79.3 → v1.81.1** (first commit) — brings the reworked tiered buffer pools (256B/4KB/16KB/32KB/1MB) and the `mem.Buffer.Slice` API; the codec-relevant interfaces are unchanged between the versions. protobuf bumps to v1.36.11 alongside.
- **`encoding.CodecV2` registration** (second commit):
  - Outgoing messages marshal via the generated `SizeVT` + `MarshalToSizedBufferVT` directly into gRPC's shared buffer pools (plain buffer below the pooling threshold, mirroring the stock codec's structure).
  - Incoming messages unmarshal straight from the transport buffer: in the common single-buffer case `MaterializeToBuffer` only takes a reference — no copy — and the buffer returns to the pool immediately after, which is safe because `UnmarshalVT` copies bytes fields out (the `unsafe` aliasing variant is deliberately not used).
  - Messages without vtproto codegen (e.g. the gRPC health service) keep the standard proto fallback.
- The V1 registration is removed — required, since gRPC's codec lookup checks the V1 registry *first*, so any V1 codec named `proto` would shadow the V2 one through the bridge. `TestVtprotoCodec_Registration` pins exactly that (asserts the V1 registry stays empty), so a future re-registration can't silently reintroduce the copy.

### Verification

- Codec tests: round-trip below/above the pooling threshold, multi-buffer reassembly, standard-proto fallback, invalid-message errors, plus the registration pin above.
- Codec benchmark (4KiB message):

  | | ns/op | B/op |
  |---|---|---|
  | Unmarshal v1-bridge | 1747 | 8984 |
  | Unmarshal v2 | **729** | **4216** |
  | Marshal v1-bridge | 728 | 4864 |
  | Marshal v2 | **268** | **16** |

  2.4x faster receive (the bridge copy is gone), 2.7x faster send with ~300x less allocation (pooled buffers).
- `go test -race` green across all five workspace modules with the upgraded grpc — every test in the repo exercises this codec. (`tests/balancer` flaked once with the known coordinator-vs-stopped-server signature being addressed in #1163; unrelated to this change.)
- `golangci-lint` clean.

Related: #1158–#1164 (performance review series). This one multiplies across every RPC: it compounds with #1162 (WAL marshal buffers) since the same write payload previously paid both costs back to back.


merlimat:
Splitting this into two independent PRs as discussed: #1168 (grpc-go upgrade to v1.81.1) and #1167 (CodecV2 registration). The codec only uses APIs identical between the two grpc versions, so they merge in any order.

PR #1196, opened 2026-06-18
fix: send assignments after dataserver handshake

mattisonchao:
## Motivation
- Data servers could receive assignment streams before completing the coordinator instance-id handshake.
- For an initial cluster config with no namespaces, the coordinator has no shard controllers to publish assignments later, so data servers could remain not ready.

## Modifications
- Return a retryable not-ready error until a data server reaches `Running` or `Draining`, with a dedicated log message for that expected startup state.
- Subscribe directly to the shard-assignment watch and send the current assignment set immediately when the stream opens.
- Add an e2e regression test that starts a coordinator with no namespaces and verifies the data server readiness probe becomes serving.

## Testing
- `go test ./oxiad/coordinator/runtime/controller`
- `go test ./oxiad/coordinator/runtime/...`
- `go test ./tests/coordinator -run TestCoordinator_NoNamespacesSendsEmptyAssignments -count=1`

### DIFF
--- a/oxiad/coordinator/runtime/controller/dataserver_controller.go
+++ b/oxiad/coordinator/runtime/controller/dataserver_controller.go
@@ -36,6 +36,7 @@ import (
 	"github.com/oxia-db/oxia/common/process"
 	"github.com/oxia-db/oxia/common/proto"
 	commontime "github.com/oxia-db/oxia/common/time"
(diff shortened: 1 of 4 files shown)


merlimat:
I think this is stil going to be a problem. #1199 only solves the part for initial leader election, not for assignments stream

PR #1204, opened 2026-07-01
feat: expose shard admin status view

mattisonchao:
## Motivation
Shard status already exists in coordinator metadata, but there is no admin API or CLI read path for operators to inspect a namespace shard set or an individual shard. This adds explicit shard status reads on top of the namespace status view PR.

## Modifications
- Add `ListShards`, `GetShard`, and `ShardView` to the admin proto.
- Populate shard views from coordinator namespace status metadata with deterministic shard ordering.
- Add `AdminClient` methods and `admin shard get` CLI support for table, name, JSON, and YAML output.
- Add unit coverage for the client, management server, and shard CLI.

## Testing
- `go test ./cmd/admin/... ./oxia ./oxiad/coordinator`
- `go test ./tests/coordinator -run TestAdminNamespaceCreateAndGet -count=1`
- `make lint`


mattisonchao:
Closing this because the namespace admin status view now exposes shard metadata through namespace_status.shards, so a separate shard admin status view is not needed.

PR #1213, opened 2026-07-07
fix: quiet assignment initialization retry

mattisonchao:
## Motivation

The coordinator can start the data server assignment stream before the data server has finished binding the coordinator instance id through `Handshake`. In that short startup window, `PushShardAssignments` can return `ErrNotInitialized`.

That condition is already retried by the existing assignment retry loop, so it should not be logged as a warning. This keeps retry behavior unchanged: canceled retries stay silent, expected initialization retries log at debug, and real assignment failures still log at warn.

The retry loop must also continue to retry the latest assignment after send errors without waiting for a new assignment update. The added test locks in that behavior.

## Modifications

- Handle assignment retry notification errors with a single `switch`.
- Keep `context.Canceled` silent.
- Log `constant.ErrNotInitialized` at debug while the retry loop continues.
- Preserve warn-level logging for other assignment dispatch failures.
- Add coverage that a send error retries the latest assignment without requiring another update.

## Testing

- `go test -count=1 ./oxiad/coordinator/runtime/controller/dataserver`
- `make lint`

### DIFF
--- a/oxiad/coordinator/runtime/controller/dataserver/dataserver_controller.go
+++ b/oxiad/coordinator/runtime/controller/dataserver/dataserver_controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/oxia-db/oxia/oxiad/coordinator/rpc"
 	controllerapi "github.com/oxia-db/oxia/oxiad/coordinator/runtime/controller"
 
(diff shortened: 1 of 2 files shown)


mattisonchao:
Closing per request; assignment-handshake changes were reverted.

PR #1246, opened 2026-07-20
perf: decode WAL entries zero-copy in ApplyLogEntry

merlimat:
### Motivation

The follower apply path decodes every committed entry through a pooled `LogEntryValue` before `ProcessWrite`. A ser/de benchmark measured the pooled `UnmarshalVT` at ~6.6µs for a 32-put/512B batch vs ~2.5µs with `UnmarshalVTUnsafe` (~2.7x), since the unsafe variant aliases string/bytes fields into the source buffer instead of copying every field.

### Changes

- **`statemachine.ApplyLogEntry` / `ApplyLogEntryWithSplitFilter`** now decode with `UnmarshalVTUnsafe`, with a comment documenting the lifetime contract. The aliasing is safe because `entry.Value` is a private heap buffer — the WAL reader materializes each `LogEntry` with a copying unmarshal, and the segment codec copies records out of the mmap — and every consumer traced through `ProcessWrite`/`ProcessControlRequest` copies what it persists (Pebble batch arena, sealed notifications) and retains nothing past the call. The pool only recycles the top-level `LogEntryValue` (`ResetVT` nils the oneof), so aliased sub-messages are never handed out again.
- **Prerequisite fix in `applyPut`**: it returned its pooled `StorageEntry` with the request's `Value`/`SecondaryIndexes` still attached. `ResetVT` retains bytes capacity across pool cycles, and the copying `Deserialize` in the delete-range paths appends into that retained capacity — under zero-copy decode that borrowed buffer is the WAL entry payload, still aliased by the remaining operations of the same entry (and the pool is process-global, so a concurrent apply on another shard could be corrupted as well). `applyPut` now detaches the borrowed buffers right after `PutMarshalable`, their last reader.
- **Regression test** (`TestApplyLogEntry_DeleteRangeDoesNotCorruptAliasedEntry`): a single entry with puts followed by delete-ranges, where the deleted key's stored value is sized to exactly fill the donated pool capacity. With the detach reverted it fails as predicted: the range-delete notification key turns into the deleted entry's stored bytes and the second range delete skips its key.
- A guard comment in `wal.readAtIndex` marking the copying `LogEntry` unmarshal as load-bearing for the zero-copy apply path.

### Verification

- `make test` (race, all modules incl. `tests/` integration): 90/90 packages ok
- Maelstrom lin-kv, 4 nodes, 60s, `--nemesis kill` (forces WAL replay through the new decode on every restart): `:valid? true`
- Maelstrom lin-kv, 4 nodes, 60s, `--nemesis partition`: `:valid? true`

A possible follow-up: `readAtIndex` could decode the `LogEntry` itself with `UnmarshalVTUnsafe` (the codec already hands it a private copy), eliding one more full-payload copy per WAL read — left out here since it widens the contract to all `Reader` consumers.

### DIFF
--- a/oxiad/dataserver/controller/statemachine/state_machine.go
+++ b/oxiad/dataserver/controller/statemachine/state_machine.go
@@ -25,7 +25,21 @@ func ApplyLogEntry(db database.DB, entry *proto.LogEntry, updateOperationCallbac
 	logEntryValue := proto.LogEntryValueFromVTPool()
 	defer logEntryValue.ReturnToVTPool()
 
-	if err := logEntryValue.UnmarshalVT(entry.Value); err != nil {
+	// UnmarshalVTUnsafe aliases every string/bytes field of the decoded tree
+	// into entry.Value instead of copying it. The lifetime contract:
+	//   - entry.Value is a private heap buffer: the WAL reader materializes
+	//     every LogEntry with a copying unmarshal (and the segment codec
+	//     itself copies records out of the mmap), so the buffer is never
+	//     mutated, recycled, or unmapped behind the aliases.
+	//   - Everything ProcessWrite/ProcessControlRequest persists is copied
+	//     (Pebble batch arena, sealed notifications) before returning, and
+	//     nothing retains the aliased strings past the call.
+	//   - The pool only recycles the top-level LogEntryValue: ResetVT nils
+	//     the oneof, so aliased sub-messages are never handed out again.
+	//   - Pooled StorageEntry objects must not keep request-owned buffers on
+	//     pool return (see applyPut), or a later pooled unmarshal would
+	//     append over this entry's payload while parts of it are still live.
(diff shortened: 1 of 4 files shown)


merlimat:
Addressed both review comments in 32d5f264:

- **Pooled entry leak on callback early-exit** (pre-existing, but worth fixing here): the pool return is now registered as a closure `defer` before the `OnPut` callback, so the entry read for the version check is recycled on the error/non-OK paths too. The closure late-binds `se`, which is required since `se` is replaced with a fresh pooled entry when the key is new — a plain `defer se.ReturnToVTPool()` at that point would capture nil.
- **Regression test pool-reuse determinism**: bumped to 10 rounds. Re-verified the negative control: against `main`'s `applyPut` (no detach) the test fails in multiple rounds; with the fix it passes. Affected packages re-run with `-race`: statemachine, database, kvstore, lead, follow — all ok.

PR #1252, opened 2026-07-25
perf: cache the last sequence values per prefix

merlimat:
### Motivation

Item 2.7 (sequences sub-item) from the performance review: every sequential put re-derived the current tail of its sequence from storage — `FindLower` opens and seeks a **fresh pebble iterator on the indexed batch** per put, `fmt.Sscanf` parses each level of the last key, and a chained `fmt.Sprintf` formats each level of the new one.

### Changes

`ProcessWrite` is single-threaded per shard, so the last sequence values per prefix are cacheable on the `db` with no locking. A cache hit skips the batch iterator and the parsing entirely; the new key is formatted with a zero-padded `strconv` append producing the same byte layout. `ErrMissingSequenceDeltas` (a put carrying fewer deltas than the sequence has levels) is enforced identically on the cached path.

The cache is invalidated wherever storage can move without it — analyzed by direction, since only one direction is dangerous:

- **Cache behind storage** (could emit a duplicate/lower key — the dangerous direction) has exactly one path: a direct client put of a sequence-shaped key (`prefix` + `-<20 digits>` segments). Invalidated in `applyPut`; the shape check is a few byte comparisons and exits immediately when the cache is empty, so plain puts pay effectively nothing.
- **Point deletes** invalidate the affected prefix and **range deletes** drop the cache, preserving today's behavior exactly: deleting the tail makes the next put re-read storage and reuse the freed value.
- **Cache ahead of storage** (emits a gap; keys stay unique — safe): session-expiry deletions of ephemeral sequence keys bypass the invalidation by design and are documented as such.
- Snapshot loads construct a fresh `db`, so no stale cache survives them; the split filter runs on that fresh instance before any sequential put.

The cache is bounded at **64K prefixes** and dropped wholesale when full. Measured footprint (Go swiss-table map, heap-alive delta): ~140–380 B per entry for realistic prefix lengths → **~9–24 MB per shard worst case**, only on shards that actually use that many sequence prefixes; an empty map costs nothing.

### Benchmark

`BenchmarkSequencePut` — full `ProcessWrite` with one sequential put (Apple M1 Max):

| | time | allocs |
|---|---|---|
| before | 8.4 µs | 65 |
| after | **4.9 µs** | **55** |

−42% per sequential put: the iterator+parse+format chain was over 40% of the whole write.

### Verification

- 7 new tests: cross-batch continuity, same-batch continuity (previously read-your-writes via the indexed batch, now via the cache), multi-level advancement, `ErrMissingSequenceDeltas` on the cached path, and the three invalidation cases. Verified to discriminate: with invalidation disabled, exactly the direct-put, delete and delete-range tests fail — the direct-put one is the duplicate-key case.
- `go test -race` green on `oxiad/dataserver/database/...` and the `tests/sequence` e2e suite.
- `golangci-lint` clean across all go.work modules on the CI-pinned v2.6.2.

### DIFF
--- a/oxiad/dataserver/database/db.go
+++ b/oxiad/dataserver/database/db.go
@@ -211,7 +211,9 @@ type db struct {
 
 	// Last sequence values per prefix key, maintained by
 	// generateUniqueKeyFromSequences. Only touched from ProcessWrite, which is
-	// single-threaded per shard, so it needs no locking.
+	// single-threaded per shard, so it needs no locking. Bounded at
+	// maxSequenceCacheEntries prefixes (dropped wholesale when full), so the
+	// worst-case footprint is that many prefix strings plus a few words each.
 	sequenceCache map[string][]uint64
 
 	putCounter                metric.Counter
--- a/oxiad/dataserver/database/db.go
+++ b/oxiad/dataserver/database/db.go
@@ -140,6 +140,7 @@ func NewDB(namespace string, shardId int64, factory kvstore.Factory,
 		notificationsEnabled:  true,
 		enabledFeatures:       sync.Map{},
 		sequenceWaiterTracker: NewSequencesWaitTracker(),
+		sequenceCache:         map[string][]uint64{},
 		log: slog.With(
 			slog.String("component", "db"),
 			slog.String("namespace", namespace),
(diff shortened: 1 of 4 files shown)


mattisonchao:
I think we should be more conservative about moving this state
outside the KV storage layer. Sequence generation depends on the
previously committed value, so changes to that state must be atomic
with the corresponding KV write.

For example, what happens if the cache is updated but the write fails
before the Pebble batch commits? The cache would be ahead of durable
storage, and retrying the operation could generate a different
sequence value.

IMO, we should keep the low-risk strconv optimisation but reconsider
the cache design. If caching is necessary, it should live in the KV
layer, where cache updates can be applied atomically with—or only after
a successful batch commit.

merlimat:
> I think we should be more conservative about moving this state
> outside the KV storage layer.

The sequence generation is implemented completely outside of kv storage layer

mattisonchao:
Yes, the logic is in the storage layer, but the state is still read from KV. 

Would it be better to move the cache inside the KV and then follow the batch commit behaviour?

PR #1253, opened 2026-07-26
fix: keep deletes when filtering a write request for a shard split

basavaraj-sm05:
When a shard splits, each child replays the parent's write-ahead log through FilterWriteRequestForSplit, which keeps only the operations that fall in the child's hash range. Puts are placed by the hash of their partition key when one is set, matching how the client routed them, but the delete loop hashes the delete's own key instead. A DeleteRequest carries no partition key, so for any record written with a partition key the delete gets hashed by a value unrelated to where the record actually lives, and it is dropped from the child that holds the record whenever the key hash lands in a different range. The put survives and the delete does not, so a key the client already deleted comes back in that child after the split and stays readable to later gets and scans. I noticed it while comparing the three placement paths: the put loop here and the snapshot-side isUserKeyInRange both honor the partition key, while this delete loop cannot, since the field is not on the message. A delete for a key a child does not hold is a no-op, so the fix keeps every delete instead of filtering it, the same way the delete ranges just below are already passed through. I also updated the existing delete test, which had encoded the old behavior, and added one that reproduces the resurrection with a partition-keyed record.

### DIFF
--- a/oxiad/dataserver/database/split_filter.go
+++ b/oxiad/dataserver/database/split_filter.go
@@ -331,13 +331,14 @@ func FilterWriteRequestForSplit(req *proto.WriteRequest, hashRange *proto.HashRa
 		}
 	}
 
-	var deletes []*proto.DeleteRequest
-	for _, d := range req.Deletes {
-		h := hash.Xxh332(d.Key)
-		if isHashInRange(h, hashRange) {
-			deletes = append(deletes, d)
-		}
-	}
+	// A DeleteRequest carries no partition key, so it can only be hashed by its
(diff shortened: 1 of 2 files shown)


mattisonchao:
It seems like the issue is caused by the client support routing the delete by 
partition key, but the server does not persist that. 

Ideally, we should carry the `PartitionKey` as part of the state in the `DeleteRequest`
to avoid this issue in future state ownership changes. 

But the current PR LGTM.

Thanks for your contribution @basavaraj-sm05 👍🏼 


Cc @merlimat

mattisonchao (approved):

PR #1258, opened 2026-07-29
fix: clamp the separator count in the hierarchical key encoder

basavaraj-sm05:
The hierarchical encoder records a key's nesting depth in a two-byte prefix, and the top bit of that prefix is what marks a key as one of oxia's own `__oxia/` keys. The depth is `strings.Count(key, "/")` over a key that arrives straight from a client Put, and it is narrowed to uint16 with nothing bounding it, so a key carrying 32768 separators sets the internal marker on its own. IsInternalKey then reports true for ordinary user data, and because newIterOptions bounds user iterations at the start of the internal region, that record stays fetchable by Get but stops appearing in List and RangeScan, and it sits above the bounds an ordinary DeleteRange would cover. Past 65535 separators the count wraps to zero, Decode takes its zero-prefix fast path and never restores the 0xff bytes back to '/', so a scan hands back a different key than the one that was written. I noticed it from the comment above InternalKeyRange, which puts the boundary at exactly 32768 separators and then leaves nothing enforcing it. hierarchical is the default sorting and nothing caps key size on the client API, so a 32 KiB key is enough to reach either case. Clamping the count just under the marker leaves every key with fewer separators encoded byte for byte as before, and only keys deeper than the prefix can count end up sharing the deepest level.

### DIFF
--- a/common/compare/encode.go
+++ b/common/compare/encode.go
@@ -60,6 +60,11 @@ const (
 	encodedSeparator      = 0xff
 	internalKeysBitMarker = 1 << 15
 
+	// maxSeparatorCount is the deepest level the 2-byte prefix can record. The
+	// count shares those bytes with the internal-key marker, so it has to stay
+	// below it.
+	maxSeparatorCount = internalKeysBitMarker - 1
+
 	// maxByteValue is the largest byte value; a prefix made only of these has
 	// no successor.
 	maxByteValue = 0xff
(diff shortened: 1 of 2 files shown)


common/compare/encode.go
```diff
@@ -97,6 +102,12 @@ func (encoderHierarchical) Encode(key string) []byte {
 			sepCount--
 		}
 
+		if sepCount > maxSeparatorCount {
+			// Keys deeper than the prefix can count all share the deepest
+			// level, rather than wrapping into the internal-key region
+			sepCount = maxSeparatorCount
```
mattisonchao, line 108: This is an edge case. We don't need to consider the compatibility issue.

mattisonchao (approved):

Issue #1132, opened 2026-05-27
Embeddable version of oxia


fenos:
Hello! I love the project, it's an absolute masterpiece! Thanks for sharing it in the open source!

It could be very nice to offer `oxia-db` as an optional embeddable library, most practically allowing importing the `dataserver` and the `coordinator` from `oxiad` and being able to instantiate it in user space.

This will allow for very interesting topologies and setups on projects using oxia-db.
What are your thoughts on this? 

Would you consider accepting a PR for this?

Thanks, 
Fabrizio

mattisonchao:
Hi @fenos 

Thanks for the contribution

Before the implementation, do you have any proposals or use cases for sharing?

fenos:
Hi @mattisonchao 

I'm prototyping a messaging broker entirely written in go and I'm using oxia-db to keep the relevant consumer groups metadata.

However, I really like the idea that oxia can run on the same binary as the broker specifically for small cluster topologies say 3-12 nodes.

For larger topologies then it would make sense to scale them independently.

the ergonomics of being able to simply embedding it and the single binary being able to communicate with other nodes it's awesome from my point of view.

It makes deployments easier for small teams (ala Kraft)

I can follow up with a proposal tomorrow

merlimat:
@fenos That's an interesting use case. It would definitely be good to see an embedded Oxia. 

It shouldn't be too difficult to pull off. Many tests are already working by doing a full-circle coordinator + data servers in a single process. 

> I'm prototyping a messaging broker entirely written in go and I'm using oxia-db to keep the relevant consumer groups metadata.

One quick note. If it's not meant to be super-high scale of traffic, take look at sequential writes + key notifications. It's essentially the building block for a pub-sub mechanism.

PR #1281, opened 2026-09-01
feat: support embedding oxia as a library (#1132)

fenos:
Closes the "how" half of #1132: make the data server and the coordinator embeddable in a Go application through the same entry points the `oxia` binary uses, so the embedding API cannot drift from the shipped server.

## What changes

- `dataserver.New(ctx, options)` and `coordinator.New(ctx, options, ...ServerOption)` take plain `Options`. Defaults are applied and the options validated before the server starts. The watch-based constructors (`dataserver.NewWithOptionsWatch`, `coordinator.NewGrpcServer`) stay for dynamic reload, and the CLI keeps using them.
- Two coordinator options for embedders:
  - `WithOnLeadershipLost(func())` replaces the default `os.Exit(1)` on metadata-leadership loss, which an embedding process cannot accept.
  - `WithInitialClusterConfiguration(*proto.ClusterConfiguration)` bootstraps the cluster configuration through the new `metadata.Factory.SeedClusterConfig`, which is idempotent and never overwrites an existing configuration.
- `UpdateOptions` on `dataserver.Server` and `coordinator.GrpcServer`; `PublicPort` and `InternalPort` accessors on the coordinator, needed when binding to ephemeral ports.
- `NewStandalone` applies option defaults and validation; `NumShards` defaults to 1.
- Provider name constants re-exported in `coordinator/option`, so choosing the metadata provider does not require importing `metadata/common`.
- Package docs on `dataserver` and `coordinator` with a full in-process cluster example, and a README section on embedding.

## Starting an embedded server

The simplest form is a standalone single-node server, the embedded equivalent of `oxia standalone`:

```go
import (
	"github.com/oxia-db/oxia/oxia"
	"github.com/oxia-db/oxia/oxiad/dataserver"
)

config := dataserver.StandaloneConfig{}
config.DataServerOptions.Server.Public.BindAddress = "localhost:0"
config.DataServerOptions.Storage.Database.Dir = "./data/db"
config.DataServerOptions.Storage.WAL.Dir = "./data/wal"

server, err := dataserver.NewStandalone(config)
if err != nil {
	return err
}
defer server.Close()

client, err := oxia.NewSyncClient(server.ServiceAddr())
```

A replicated in-process cluster runs a coordinator next to the data servers; the `coordinator` package documentation walks through a three-node example bootstrapped with `WithInitialClusterConfiguration`.

## Tests

- `tests/embedded`: a coordinator plus three data servers in one process through the public API, exercised with the client, plus the standalone form above.
- Unit tests for `SeedClusterConfig` (seeds once, never overwrites, rejects a nil configuration) and for the server options (a nil option or nil handler falls back to the default).
- Existing e2e/mocks/security tests moved to the new constructors where they built servers by hand.

## Context

The use case is a broker written in Go that keeps its metadata in Oxia and wants a single binary for small topologies (3 to 12 nodes), KRaft-style, while larger deployments run Oxia separately. This PR is the minimal library surface for that. Follow-ups I'd like to propose separately, each small and useful on its own:

1. namespaces (with per-namespace key sorting) in `NewStandalone`, allocating shard IDs the way the coordinator does, so a standalone data directory is a valid cluster layout;
2. seeding cluster *status* from a server's existing shards on first coordinated start, the twin of `SeedClusterConfig`;
3. raft metadata provider: membership reconciled to a peer list (`AddVoter`/`RemoveServer`) and a real leader hint under raft;
4. growing an ensemble to a raised `replication_factor`.

Happy to split or reshape any of this to match how you'd like the surface to look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,27 @@ With Oxia, you get a scalable, robust, and flexible solution for managing metada
 
 <br>
 
+### Embedding Oxia
+
+Besides running as dedicated `oxia server` / `oxia coordinator` processes, Oxia can be embedded as a library inside a Go application, so that each application node hosts an Oxia data server (and, on some nodes, the coordinator) in the same binary:
+
+```go
+import (
+    "github.com/oxia-db/oxia/oxiad/dataserver"
+    "github.com/oxia-db/oxia/oxiad/dataserver/option"
+)
+
+options := option.NewDefaultOptions()
+options.Storage.Database.Dir = "./data/db"
+options.Storage.WAL.Dir = "./data/wal"
+
+server, err := dataserver.New(ctx, options)
+```
+
+These are the same entry points the `oxia` binary itself is built on. See the [`dataserver`](oxiad/dataserver/doc.go) package documentation for embedding a data server or a standalone single-node server, and the [`coordinator`](oxiad/coordinator/doc.go) package documentation for bootstrapping a whole cluster in-process with `coordinator.New`.
+
+<br>
+
 ### Contributing to Oxia
 
 Please 🌟 star the project if you like it. 
--- a/cmd/server/cmd.go
+++ b/cmd/server/cmd.go
@@ -147,6 +147,6 @@ func exec(cmd *cobra.Command, _ []string) {
 			}
 		}
 
-		return dataserver.New(context.Background(), optionsWatch)
+		return dataserver.NewWithOptionsWatch(context.Background(), optionsWatch)
 	})
 }
--- a/oxiad/coordinator/doc.go
+++ b/oxiad/coordinator/doc.go
@@ -1,0 +1,71 @@
+// Copyright 2023-2026 The Oxia Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package coordinator implements the Oxia coordinator: it manages the
+// cluster status, assigns shards to data servers and drives leader
+// elections. It backs the `oxia coordinator` command and can equally be
+// embedded in a Go application together with the dataserver package, running
+// a whole Oxia cluster in-process:
+//
+//	ctx := context.Background()
+//
+//	var identities []*proto.DataServerIdentity
+//	for i := 0; i < 3; i++ {
+//		dsOptions := dsoption.NewDefaultOptions()
+//		dsOptions.Server.Public.BindAddress = "localhost:0"
+//		dsOptions.Server.Internal.BindAddress = "localhost:0"
+//		dsOptions.Storage.Database.Dir = filepath.Join(dataDir, strconv.Itoa(i), "db")
+//		dsOptions.Storage.WAL.Dir = filepath.Join(dataDir, strconv.Itoa(i), "wal")
+//
+//		server, err := dataserver.New(ctx, dsOptions)
+//		if err != nil {
+//			return err
+//		}
+//		defer server.Close()
+//
+//		identities = append(identities, &proto.DataServerIdentity{
+//			Public:   fmt.Sprintf("localhost:%d", server.PublicPort()),
+//			Internal: fmt.Sprintf("localhost:%d", server.InternalPort()),
+//		})
+//	}
+//
+//	options := option.NewDefaultOptions()
+//	options.Server.Public.BindAddress = "localhost:0"
+//	options.Server.Internal.BindAddress = "localhost:0"
+//	options.Metadata.ProviderName = option.ProviderMemory
+//
+//	coord, err := coordinator.New(ctx, options,
+//		coordinator.WithInitialClusterConfiguration(&proto.ClusterConfiguration{
+//			Namespaces: []*proto.Namespace{{
+//				Name:              constant.DefaultNamespace,
+//				ReplicationFactor: 3,
+//				InitialShardCount: 1,
+//			}},
+//			Servers: identities,
+//		}))
+//	if err != nil {
+//		return err
+//	}
+//	defer coord.Close()
+//
+//	client, err := oxia.NewSyncClient(identities[0].Public)
+//
+// The example uses the in-memory metadata provider, which keeps the cluster
+// status inside the coordinator process and is suitable when a single
+// coordinator is embedded. Multi-coordinator topologies should use the file
+// or raft metadata providers instead, and must decide what happens when a
+// coordinator loses the metadata leadership: by default the whole process
+// exits, which embedding applications usually want to override with
+// [WithOnLeadershipLost].
+package coordinator
--- a/oxiad/coordinator/grpc_server.go
+++ b/oxiad/coordinator/grpc_server.go
@@ -54,7 +54,9 @@ type serverOptions struct {
 func newServerOptions(serverOpts []ServerOption) serverOptions {
 	so := serverOptions{onLeadershipLost: onLeadershipLost}
 	for _, opt := range serverOpts {
-		opt(&so)
+		if opt != nil {
+			opt(&so)
+		}
 	}
 	return so
 }
@@ -77,6 +79,11 @@ func (so *serverOptions) seedClusterConfig(metadataFactory *coordmetadata.Factor
 // synchronously; spawn a goroutine instead.
 func WithOnLeadershipLost(handler func()) ServerOption {
 	return func(so *serverOptions) {
+		if handler == nil {
+			// A nil handler would panic at the worst moment; keep the
+			// fail-safe default instead.
+			handler = onLeadershipLost
+		}
 		so.onLeadershipLost = handler
 	}
 }
--- a/oxiad/coordinator/grpc_server.go
+++ b/oxiad/coordinator/grpc_server.go
@@ -16,6 +16,7 @@ package coordinator
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"os"
 	"sync"
@@ -42,6 +43,55 @@ import (
 	commonrpc "github.com/oxia-db/oxia/oxiad/common/rpc"
 )
 
+// ServerOption configures optional behavior of the coordinator server.
+type ServerOption func(*serverOptions)
+
+type serverOptions struct {
+	onLeadershipLost     func()
+	initialClusterConfig *proto.ClusterConfiguration
+}
+
+func newServerOptions(serverOpts []ServerOption) serverOptions {
+	so := serverOptions{onLeadershipLost: onLeadershipLost}
+	for _, opt := range serverOpts {
+		opt(&so)
+	}
+	return so
+}
+
+func (so *serverOptions) seedClusterConfig(metadataFactory *coordmetadata.Factory) error {
+	if so.initialClusterConfig == nil {
+		return nil
+	}
+	return metadataFactory.SeedClusterConfig(so.initialClusterConfig)
+}
+
+// WithOnLeadershipLost overrides what happens when this coordinator loses the
+// metadata leadership. The default handler terminates the process, which is
+// the right behavior for a dedicated coordinator process but not when the
+// coordinator is embedded in a larger application.
+//
+// A coordinator that lost the leadership must stop coordinating: the handler
+// is expected to arrange for the server to be closed. It is invoked from an
+// internal goroutine that Close waits on, so the handler must not call Close
+// synchronously; spawn a goroutine instead.
+func WithOnLeadershipLost(handler func()) ServerOption {
+	return func(so *serverOptions) {
+		so.onLeadershipLost = handler
+	}
+}
+
+// WithInitialClusterConfiguration seeds the metadata store with the given
+// cluster configuration when none exists yet. It allows bootstrapping a
+// cluster programmatically, without a cluster configuration file or a
+// separate admin call. If a configuration is already present it is left
+// untouched.
+func WithInitialClusterConfiguration(config *proto.ClusterConfiguration) ServerOption {
+	return func(so *serverOptions) {
+		so.initialClusterConfig = config
+	}
+}
+
 type GrpcServer struct {
 	// concurrent control
 	ctx          context.Context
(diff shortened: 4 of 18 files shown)


fenos:

oxiad/coordinator/grpc_server.go
```diff
@@ -42,6 +43,55 @@ import (
...
+//
+// A coordinator that lost the leadership must stop coordinating: the handler
+// is expected to arrange for the server to be closed. It is invoked from an
+// internal goroutine that Close waits on, so the handler must not call Close
+// synchronously; spawn a goroutine instead.
+func WithOnLeadershipLost(handler func()) ServerOption {
+	return func(so *serverOptions) {
+		so.onLeadershipLost = handler
+	}
+}
```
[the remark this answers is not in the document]
fenos: Addressed in b93c06af: a nil handler now keeps the fail-safe default instead of replacing it.

oxiad/coordinator/metadata/factory.go
```diff
@@ -106,6 +106,30 @@ func New(ctx context.Context, options *option.Options) (*Factory, error) {
 	return factory, nil
 }
 
+// SeedClusterConfig stores the given cluster configuration if none exists yet.
+// It is a no-op when a configuration is already present, and it tolerates
+// losing the seeding race to another coordinator.
+func (f *Factory) SeedClusterConfig(config *commonproto.ClusterConfiguration) error {
+	f.mu.Lock()
+	configProvider := f.configProvider
+	f.mu.Unlock()
```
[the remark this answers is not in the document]
fenos: Addressed in b93c06af: SeedClusterConfig now returns an error for a nil configuration, and the store is left untouched.

oxiad/coordinator/grpc_server.go
```diff
@@ -42,6 +43,55 @@ import (
...
+type serverOptions struct {
+	onLeadershipLost     func()
+	initialClusterConfig *proto.ClusterConfiguration
+}
+
+func newServerOptions(serverOpts []ServerOption) serverOptions {
+	so := serverOptions{onLeadershipLost: onLeadershipLost}
+	for _, opt := range serverOpts {
+		opt(&so)
+	}
```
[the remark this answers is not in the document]
fenos: Addressed in b93c06af: nil options are now skipped.

oxiad/coordinator/metadata/factory.go
```diff
@@ -106,6 +106,33 @@ func New(ctx context.Context, options *option.Options) (*Factory, error) {
...
+		Version: metadatacommon.NotExists,
+	}); err != nil {
+		if errors.Is(err, metadatacommon.ErrBadVersion) {
+			// Another coordinator stored a configuration first.
+			return nil
+		}
+		return err
+	}
+	return nil
+}
```
merlimat, line 134: The seeded configuration is stored without `Validate()`. The admin write path validates it in `computeConfig` (`metadata.go`), but here an invalid config goes straight into the metadata store, and it can take the host process down. For example, seeding `ReplicationFactor: 3` with a single server (stack abridged):

```
panic: unexpected behaviour
  runtime/balancer/selector/single/selector.go:44     (*server).Select
  runtime/balancer/selector/ensemble/selector.go:47   (*ensemble).Select
  runtime/runtime.go:363                              selectNewEnsemble
  runtime/runtime.go:240                              CreateNamespace
  reconciler/namespace_reconciler.go:39               Reconcile
  reconciler/cluster_reconciler.go:61                 reconciler.New
  grpc_server.go:226                                  NewGrpcServer
```

This happens on the goroutine that calls `coordinator.New`, so the embedding application crashes. With the file or raft provider, the config is written before the panic, and `SeedClusterConfig` never overwrites an existing one. Every restart then panics the same way until someone cleans up the metadata by hand.

Could you call `config.Validate()` here before storing it? It would also be good to check the option in `NewGrpcServer` before binding the ports and waiting for leadership, so a standby coordinator fails fast too.

oxiad/coordinator/metadata/factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	second := &commonproto.ClusterConfiguration{
+		Namespaces: []*commonproto.Namespace{{Name: "second"}},
+	}
+	require.NoError(t, factory.SeedClusterConfig(second))
+
+	current := factory.configProvider.Watch().Load()
+	assert.Equal(t, version, current.Version)
+	require.Len(t, current.Value.Namespaces, 1)
+	assert.Equal(t, "first", current.Value.Namespaces[0].Name)
```
merlimat, line 88: Once seeding validates, these configs (RF=0, `InitialShardCount`=0, no servers) will be rejected. Please switch them to valid ones, and add a case asserting that an invalid config (e.g. RF greater than the number of servers) is rejected and nothing is stored.

oxiad/coordinator/grpc_server.go
```diff
@@ -42,6 +43,62 @@ import (
...
+
+// WithOnLeadershipLost overrides what happens when this coordinator loses the
+// metadata leadership. The default handler terminates the process, which is
+// the right behavior for a dedicated coordinator process but not when the
+// coordinator is embedded in a larger application.
+//
+// A coordinator that lost the leadership must stop coordinating: the handler
+// is expected to arrange for the server to be closed. It is invoked from an
+// internal goroutine that Close waits on, so the handler must not call Close
+// synchronously; spawn a goroutine instead.
```
merlimat, line 79: I tried the path this comment recommends, closing the server after the leadership is lost, and `Close()` can hang forever.

After a leadership loss, every metadata status write fails (raft `VerifyLeader`, configmap conflicts). If an election is in flight, the shard controller's goroutine sits in `UpdateShardStatus`. That call retries with a backoff that never gives up, and it only stops when the metadata context is cancelled in `metadata.Close()`. But `GrpcServer.Close()` calls `runtime.Close()` first, and that waits for the same goroutine. I reproduced it by making the file provider's `cluster-status.json` read-only and stopping the data servers:

```
GrpcServer.Close → runtime.Close → shard.(*controller).Close → sync.WaitGroup.Wait
  waiting on: controller.run → Election.start → coordinatorMetadata.UpdateShardStatus → backoff.RetryNotify
```

The ordering bug predates this PR; `os.Exit` was hiding it. I'll fix that part on main.

The part that belongs to this PR: with a custom handler, the coordinator keeps running elections and balancing until the embedder's `Close` completes. The default `os.Exit` handler exists to close exactly that window (see the comment on `onLeadershipLost`). With the configmap provider, the stale leader's next status write gets `ErrBadVersion`, which `computeStatus` treats as an invariant violation and panics on. That one is from reading the code; I haven't reproduced it.

Once `Close` is fixed, I'd rather the coordinator stop coordinating by itself on loss: stop metadata writes, close the reconciler and the runtime, and set health to NOT_SERVING. Only then should it call the handler, which becomes a notification. A test that drives the loss path would be great too; today only the option plumbing is covered.

oxiad/coordinator/grpc_server.go
```diff
@@ -60,7 +117,32 @@ type GrpcServer struct {
...
+// coordinator starts. If options is nil, the default options are used.
+//
+// The coordinator keeps a reference to options: the caller must not mutate
+// them after this call and should use UpdateOptions instead.
+//
+// With a multi-coordinator metadata provider (file, configmap or raft), this
+// call blocks until the coordinator acquires the metadata leadership.
+func New(parent context.Context, options *option.Options, serverOpts ...ServerOption) (*GrpcServer, error) {
+	if options == nil {
+		options = option.NewDefaultOptions()
```
merlimat, line 131: Two small things:

- `New(ctx, nil)` always fails. `option.NewDefaultOptions()` doesn't set `Metadata.ProviderName`, so `Validate()` returns `must be one of "memory", "configmap", "raft" or "file"`. I'd drop the nil handling and the "default options are used" sentence, or pick a default provider.
- With the file or raft provider, a standby coordinator blocks in `New` and ignores `parent` cancellation. Its internal and public gRPC ports are already bound, and the caller has no handle to close. I checked with two file-provider coordinators sharing a directory: the second `New` was still blocked 5s after its context was cancelled. In the KRaft-style topology you describe, every non-leader node is in this state. Honoring `parent.Done()` while waiting (release what was started and return `ctx.Err()`) can be a follow-up, but it's worth a sentence in the doc now.

oxiad/coordinator/grpc_server.go
```diff
@@ -183,14 +268,42 @@ func NewGrpcServer(parent context.Context, optionsWatch *commonwatch.Watch[*opti
...
+// the process: a coordinator that lost the leadership must stop coordinating
+// immediately, before it can run elections or move ensembles alongside the
+// new leader, and a restart rejoins the election from a clean state.
+// Overridable in tests.
 var onLeadershipLost = func() {
 	os.Exit(1)
 }
 
+// InternalPort returns the port of the internal gRPC server used for
+// coordinator-to-coordinator communication.
```
merlimat, line 281: Nit: the coordinator's internal server only registers the health service (see `StartGrpcServer("coordinator", ...)` above). No coordinator-to-coordinator traffic goes through it; raft uses `Metadata.Raft.Address`. Maybe "the internal gRPC server (health checks)".

oxiad/coordinator/grpc_server.go
```diff
@@ -183,14 +268,42 @@ func NewGrpcServer(parent context.Context, optionsWatch *commonwatch.Watch[*opti
...
+// PublicPort returns the port of the public gRPC server exposing the
+// management (admin) API.
+func (s *GrpcServer) PublicPort() int {
+	return s.managementServer.Port()
+}
+
+// UpdateOptions publishes a new configuration to the running coordinator.
+// Only the dynamic settings (currently the log options) take effect at
+// runtime; the remaining settings require a restart.
+func (s *GrpcServer) UpdateOptions(options *option.Options) error {
```
merlimat, line 295: Same as the data server's `UpdateOptions`, see my comment there.

oxiad/dataserver/server.go
```diff
@@ -170,6 +192,21 @@ func (s *Server) InternalPort() int {
 	return s.internalRpcServer.grpcServer.Port()
 }
 
+// UpdateOptions publishes a new configuration to the running server. Only the
+// dynamic settings (currently the log options) take effect at runtime; the
+// remaining settings require a restart.
+func (s *Server) UpdateOptions(options *option.Options) error {
```
merlimat, line 198: I'd drop `UpdateOptions`, here and on the coordinator, from this PR:

- Its only runtime effect is `logging.ReconfigureLogger`, which sets the package-global `logging.LogLevel`/`logLevelVar`. That only changes anything if the host process installed Oxia's handler through `logging.ConfigureLogger()`, which calls `slog.SetDefault`. Otherwise the call is a no-op. When it does act, it acts on the whole process rather than on this server.
- With several servers in one process it's a data race. Calling `UpdateOptions` concurrently on two embedded data servers trips `-race` on `logging.LogLevel` (`logger.go:67` vs `:70`), between the two config-watcher goroutines.

Anyone who needs dynamic reload can still use `NewWithOptionsWatch` / `NewGrpcServer`. The embedded test only checks the returned error, so it wouldn't catch either issue.

oxiad/coordinator/doc.go
```diff
@@ -0,0 +1,71 @@
...
+//
+//	ctx := context.Background()
+//
+//	var identities []*proto.DataServerIdentity
+//	for i := 0; i < 3; i++ {
+//		dsOptions := dsoption.NewDefaultOptions()
+//		dsOptions.Server.Public.BindAddress = "localhost:0"
+//		dsOptions.Server.Internal.BindAddress = "localhost:0"
+//		dsOptions.Storage.Database.Dir = filepath.Join(dataDir, strconv.Itoa(i), "db")
+//		dsOptions.Storage.WAL.Dir = filepath.Join(dataDir, strconv.Itoa(i), "wal")
```
merlimat, line 29: This example fails as written. Metrics are enabled by default on `0.0.0.0:8080`, so the second data server fails with `listen tcp 0.0.0.0:8080: bind: address already in use`, and the coordinator would collide too. The tests disable metrics everywhere, which hides it. Setting `Observability.Metric.BindAddress = "localhost:0"` on each server, or disabling metrics, fixes it.

It may also be worth noting that in-process servers share one metrics pipeline: `oxiad/common/metric` installs the global OTel `MeterProvider` in `init()`, and every endpoint serves the default Prometheus registry.

oxiad/coordinator/doc.go
```diff
@@ -0,0 +1,71 @@
...
+//
+//	client, err := oxia.NewSyncClient(identities[0].Public)
+//
+// The example uses the in-memory metadata provider, which keeps the cluster
+// status inside the coordinator process and is suitable when a single
+// coordinator is embedded. Multi-coordinator topologies should use the file
+// or raft metadata providers instead, and must decide what happens when a
+// coordinator loses the metadata leadership: by default the whole process
+// exits, which embedding applications usually want to override with
+// [WithOnLeadershipLost].
```
merlimat, line 70: With this guidance, the cluster doesn't survive a restart of the embedding process:

- **Memory provider + on-disk data directories** (this example): I restarted everything with the same directories and ports. The new coordinator generates a new cluster instance id, but the data servers are still bound to the old one. Every fence then fails with `PermissionDenied: unexpected instance id`, elections retry forever, and clients get `server not initialized yet`. The data can't be reached until the directories are wiped. The memory provider only fits when the data servers' storage is ephemeral too, e.g. in tests.
- **File provider**: the same restart works, and reads return the pre-restart data. I'd recommend it for a single embedded coordinator with persistent data.
- *"Multi-coordinator topologies should use the file or raft metadata providers"*: file-provider leadership is a local file lock. Coordinators on different machines, each with its own local file, would all consider themselves leader. Across machines it needs to be raft, or configmap on Kubernetes.
- Only the raft and configmap providers can report a leadership loss; memory and file hold it for the lifetime of the process. So `WithOnLeadershipLost` only matters for those two.

oxiad/dataserver/doc.go
```diff
@@ -0,0 +1,37 @@
...
+//	}
+//	defer server.Close()
+//
+// A data server is passive until a coordinator assigns shards to it; see the
+// coordinator package for how to run a whole Oxia cluster in-process.
+//
+// [NewStandalone] starts a self-contained single-node server instead, with no
+// replication and no coordinator: the embedded equivalent of the
+// `oxia standalone` command, and the simplest option for tests and
+// single-process deployments.
```
merlimat, line 36: Small mismatch: a zero `StandaloneConfig{}` (as in the PR description) isn't quite `oxia standalone`. `NotificationsEnabled` is `false` there, while the CLI flag and `NewTestConfig` default to `true`, so clients get `oxia: notifications disabled`. Since it's a plain `bool`, `NewStandalone` can't default it. Maybe set it in the example, or mention it in the doc.

cmd/server/cmd.go
```diff
@@ -147,6 +147,6 @@ func exec(cmd *cobra.Command, _ []string) {
 			}
 		}
 
-		return dataserver.New(context.Background(), optionsWatch)
+		return dataserver.NewWithOptionsWatch(context.Background(), optionsWatch)
```
merlimat, line 150: Not a blocker, just a naming nit: the data server's watch-based constructor becomes `NewWithOptionsWatch`, while the coordinator's stays `NewGrpcServer`. Also, `dataserver.New` changes signature, so existing callers break at compile time. That's fine at this stage; I'll mention it in the release notes.

README.md
```diff
@@ -30,6 +30,27 @@ With Oxia, you get a scalable, robust, and flexible solution for managing metada
 
 <br>
 
+### Embedding Oxia
```
merlimat, line 33: It would help to have a short "limitations" note here or in the package docs. Embedded servers share process-global state and can still terminate the host:

- `oxiad/common/metric` replaces the global OTel `MeterProvider` in `init()`.
- The log level is package-global.
- `os.Exit(1)` remains on gRPC or metrics serve failures and on Pebble `Fatalf`.

I'm not asking to change those here, but embedders should know about them.

merlimat (changes requested):
Thanks @fenos! This is a nice, small surface, and reusing the same entry points as the binary is the right call. The shape looks good to me: a plain-options `New` next to the watch-based constructors the CLI keeps using, functional options on the coordinator, an idempotent config seed, and tests that go through the public API.

I tried it in a few scenarios an embedder will run into: a bad config, restarts, standby coordinators, and leadership loss. A few things need fixing before we merge. Details and repro steps are in the inline comments. In short:

1. **The seeded config isn't validated.** An invalid one panics the host process inside `coordinator.New`. With the file or raft provider it also persists, so every restart panics again.
2. **The provider guidance in the docs.** Memory provider + on-disk data doesn't survive a restart of the embedding process. The file provider isn't safe for coordinators on different machines.
3. **The `coordinator` package example fails as written**: every server tries to bind metrics on port 8080.
4. **`UpdateOptions`**: I'd drop it from this PR. It has no effect for a typical embedder and races when several servers run in one process.
5. **`WithOnLeadershipLost`**: the coordinator should stop coordinating by itself before calling the handler. Separately, closing it after a leadership loss can currently hang. That's a pre-existing ordering bug in `Close()` that `os.Exit` was hiding; I'll fix it on main.
6. **DCO**: 7f5c233 and b93c06a are missing a `Signed-off-by`. Rebasing onto current main with `--signoff` (or squashing into a single signed-off commit) and force-pushing will clear the check.

There are also a few smaller things inline: `New(ctx, nil)`, a standby `New` that ignores its context, the standalone notification default, and a suggested "limitations" note.

On the follow-ups you listed: happy to take them one at a time. For (2), note that every data server persists the cluster instance id it was bound to (in its manifest), and it rejects a coordinator whose id doesn't match. That's what breaks the memory-provider restart described inline, so a seeded status would need to carry the same instance id.

PR #1282, opened 2026-09-01
fix: an inverted range scans nothing instead of tripping pebble's bound invariant

fenos:
`RangeScan`, `KeyRangeScanReverse` and the batch `RangeScan` pass the caller's bounds straight into `pebble.IterOptions`. When the range is inverted or empty — lower bound at or past the upper bound — pebble's behaviour is undefined: release builds happen to produce an empty iterator, but invariant builds (which the race detector switches on) assert with

```
WARN mergingIter: lower bound violation: <lower> < <upper>
```

inside the leader's apply path. The everyday trigger is a `DeleteRange` whose `start_inclusive >= end_exclusive`; a client expecting a no-op gets a crashed data server under `-race`.

## Fix

Judge the range before pebble sees it. Encoded keys order bytewise, so a lower bound at or past the upper bound returns an iterator that is never valid. Applied to the three bounded constructors; single-bound iterators are unaffected.

## Test

`TestPebbleRangeScanInvertedBoundsIsEmpty` covers the store scan, the reverse scan and the batch scan for both an inverted and an empty range. It fails on `main` under `go test -race` with the assertion above and passes with the fix.

https://claude.ai/code/session_017zeKZ9GicpAvatkAh1mutZ

### DIFF
--- a/oxiad/dataserver/database/kvstore/kv_pebble.go
+++ b/oxiad/dataserver/database/kvstore/kv_pebble.go
@@ -536,6 +536,9 @@ func (p *Pebble) KeyRangeScanReverse(lowerBound, upperBound string, itOpts Itera
 	if upperBound != "" {
 		ub = p.keyEncoder.Encode(upperBound)
 	}
+	if invertedRange(lb, ub) {
+		return emptyIterator{}, nil
+	}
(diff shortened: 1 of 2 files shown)


fenos:

PR #1283, opened 2026-09-01
fix: order a multi-shard index scan by secondary key

fenos:
A `RangeScan` with `UseIndex` fans out to every shard and merges the per-shard streams through `ResultHeap`, whose `Less` compares **record keys** only. Each shard returns its slice in secondary-key order, so the merge interleaves them by the wrong key: an index scan spanning shards comes back out of secondary-key order, while the same scan on a one-shard namespace is sorted. `compareGetResponse` already applies the right rule (secondary key first, then record key) to multi-shard gets; the scan merge never used it — and could not have, because the server's index range-scan iterator dropped the secondary key while parsing the composite entry (`secondaryIndexGet` does report it).

## Fix

Server: `secondaryIndexRangeIterator.Value()` sets `secondary_index_key` from the composite index entry, as the index `Get` path already does.

Client: `GetResult` gains `SecondaryIndexKey`, filled from the response (empty for plain reads), and `ResultHeap.Less` orders by it when both results carry one, ties by record key; plain scans keep record-key order.

## Tests

- `TestSecondaryIndices_RangeScanReturnsSecondaryKey` (server): each result of an index range scan carries its secondary key.
- `TestAggregateRangeScanAcrossShardsOrdersByIndexKey` (client): two shard streams whose record-key order is the reverse of their index order merge in index order; `…OrdersByKey` pins the plain-scan behaviour.

Found by running a multi-shard namespace in-process under a kv conformance suite that had only ever run against `oxia standalone --shards 1`.

https://claude.ai/code/session_017zeKZ9GicpAvatkAh1mutZ

### DIFF
--- a/oxia/client.go
+++ b/oxia/client.go
@@ -240,6 +240,11 @@ type GetResult struct {
 	// The version information
 	Version Version
 
+	// SecondaryIndexKey is the record's key in the secondary index the
+	// operation read through, when it used one — the order a RangeScan
+	// over an index follows across shards. Empty otherwise.
+	SecondaryIndexKey string
+
 	// The error if the `Get` operation failed
 	Err error
 }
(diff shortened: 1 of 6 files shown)


fenos:

PR #1291, opened 2026-09-07
feat: balance shards and leaders per namespace

### LINKED ISSUE #776: feat: load balance shards by namespace
### Motivation

The current shards' load balance would not work well if the user has multiple namespaces and most traffic is in one of the namespaces.  We should consider doing namespace-level shards to balance all the nodes by default to improve this case.

108474839:
Balance shard placement and leader selection per namespace instead of globally, preventing namespace-level imbalance from being hidden by cluster-wide totals.
Includes regression tests for namespace isolation and leader balancing.

Fixes #776


108474839:
Would it make sense to discuss load-aware balancing (e.g. shard size or traffic) and migration limits in a follow-up issue? This PR keeps the existing count-based policy and applies it per namespace.

PR #1293, opened 2026-09-10
fix: publish restored assignments on coordinator restart

### LINKED ISSUE #1292: fix: publish restored shard assignments after coordinator restart in 0.16
## Motivation

Oxia 0.16.x can leave a newly started data server permanently not ready after the active coordinator has restarted while the persisted shard state is already in steady state.

During coordinator startup, the persisted shard state is restored and the ensemble is checked. If every replica reports the expected role and term, no leader election is triggered. In this path, the coordinator assignment snapshot remains nil. The data-server assignment dispatcher opens the assignment stream and waits for a value different from its nil current value, but the coordinator value is also nil, so it can wait indefinitely.

If a data server restarts later, shard replication can still recover it as a follower through the separate replication RPC, while the data server never receives its first routing assignment and its readiness service remains NOT_SERVING.

### Reproduction

1. Start an Oxia 0.16.x cluster with three data servers, one namespace, one shard, and replication factor three.
2. Wait until the shard is in steady state with one leader and two followers.
3. Restart only the active coordinator while all data servers remain available.
4. Confirm that startup verifies the existing ensemble and does not run a new leader election.
5. Restart one follower data server.
6. Observe that replication restores the follower, but the assignment stream does not deliver its initial snapshot and readiness remains NOT_SERVING.
7. A coordinator goroutine remains in WaitForNextUpdate with a nil current assignment, while the data server waits for its first assignment message.

Source review indicates that v0.16.8 and v0.16.9 retain the same behavior. v0.16.10 initializes assignments only when there are no namespaces, so a restored non-empty namespace remains affected.

The 0.17 line addresses this through #1215 and #1216: startup reconciliation recomputes assignments, assignment state uses a snapshot/watch, each new stream sends the latest snapshot even when unchanged, and server-side stream termination is detected so the stream can reconnect.

## Modifications

- Backport the reliable initial assignment publication and stream reconnection behavior from #1215 and #1216 to release-0.16.
- At minimum, compute and publish assignments after restoring persisted non-empty shard state during coordinator startup.
- Ensure every newly established or re-established assignment stream sends the latest snapshot even when no assignment change occurred.
- Add a regression test covering a coordinator restart followed by a data-server restart with an unchanged, steady-state shard.

## Testing

- Reproduced with v0.16.7 using a generic three-data-server, one-shard topology.
- Reviewed v0.16.8, v0.16.9, v0.16.10, and the release-0.16 branch for the affected initialization and dispatch paths.
- Verified that v0.17.0 and v0.17.1 contain the snapshot subscription, initial reconciliation, reconnect delivery behavior, and regression coverage from #1215 and #1216.

mattisonchao:
## Motivation

On Oxia 0.16, a coordinator restart restores persisted shard status and verifies the existing ensemble. When every replica already reports the expected role and term, no leader election runs. Since assignment publication was tied to election and change callbacks, the restarted coordinator could retain a nil assignment snapshot indefinitely.

A data server starting afterward would connect to the assignment stream but never receive its first snapshot, leaving its readiness service in NOT_SERVING even though shard replication could recover independently.

The CI workflow also installed `go-license@latest`. Version 1.50.0 now requires Go 1.27, while release-0.16 intentionally remains on Go 1.26.

Fixes #1292.

## Modifications

- Publish the current assignment snapshot immediately after the coordinator applies the initial cluster configuration to persisted status.
- Preserve the existing leader-election updates while covering steady-state coordinator restarts and empty namespace configurations.
- Add an end-to-end regression that establishes a shard, restarts the coordinator with the persisted steady state, and verifies that the restored assignment is available immediately.
- Allow the control integration test to observe the valid transitional assignment with no leader by waiting for the shard to reach steady state before dereferencing its leader.
- Pin `go-license` to v1.49.0, the last release compatible with Go 1.26.

### DIFF
--- a/.github/workflows/pr_build_and_test.yaml
+++ b/.github/workflows/pr_build_and_test.yaml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Installing Go dependencies
         run: |
-          go install github.com/palantir/go-license@latest
+          go install github.com/palantir/go-license@v1.49.0
           go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.6.2
 
       - run: make license-check
--- a/oxiad/coordinator/coordinator.go
+++ b/oxiad/coordinator/coordinator.go
@@ -795,13 +795,12 @@ func NewCoordinator(meta metadata.Provider,
 		c.Info("Checking cluster config", slog.Any("clusterConfig", clusterConfig))
 	}
 	clusterStatus, _, _ = c.statusResource.ApplyChanges(clusterConfig, c.selectNewEnsemble)
-	if len(clusterStatus.Namespaces) == 0 {
-		// No shard controllers will run leader election and publish assignments,
-		// so send the empty assignment set now to make data servers ready.
-		c.Lock()
-		c.computeNewAssignments()
-		c.Unlock()
-	}
+	// Publish the current snapshot after restoring and reconciling status. Shards
+	// that are already healthy do not run a new leader election, so there might
+	// otherwise be no callback to initialize assignments on coordinator restart.
+	c.Lock()
+	c.computeNewAssignments()
+	c.Unlock()
 
 	// init shard controller
 	for ns, shards := range clusterStatus.Namespaces {
--- a/tests/control/control_request_test.go
+++ b/tests/control/control_request_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/oxia-db/oxia/common/proto"
 	clientrpc "github.com/oxia-db/oxia/common/rpc"
@@ -63,6 +64,11 @@ func TestControlRequestFeatureEnabled(t *testing.T) {
 	client, err := oxia.NewSyncClient(sa1.Public, oxia.WithNamespace("default"))
 	assert.NoError(t, err)
 	defer client.Close()
+
+	require.Eventually(t, func() bool {
+		shard := coordinatorInstance.StatusResource().Load().Namespaces["default"].Shards[0]
+		return shard.Status == model.ShardStatusSteadyState && shard.Leader != nil
+	}, 10*time.Second, 100*time.Millisecond)
 
 	resource := coordinatorInstance.StatusResource().Load()
 	shardMetadata := resource.Namespaces["default"].Shards[0]
--- a/tests/coordinator/coordinator_e2e_test.go
+++ b/tests/coordinator/coordinator_e2e_test.go
@@ -472,6 +472,51 @@ func TestCoordinator_NoNamespacesSendsEmptyAssignments(t *testing.T) {
 	}, 10*time.Second, 10*time.Millisecond)
 }
 
+func TestCoordinator_RestartPublishesRestoredAssignments(t *testing.T) {
+	s1, sa1 := newServer(t)
+	defer s1.Close()
+
+	metadataProvider := metadata.NewMetadataProviderMemory()
+	clusterConfig := model.ClusterConfig{
+		Namespaces: []model.NamespaceConfig{{
+			Name:              constant.DefaultNamespace,
+			ReplicationFactor: 1,
+			InitialShardCount: 1,
+		}},
+		Servers: []model.Server{sa1},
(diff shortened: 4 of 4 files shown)


merlimat:
This is for the release-0.16 branch. In reality, I'd prefer all fixes to go into `main` first and only then get cherry-picked into maintenance branches.

merlimat (commented):
The core change looks correct to me. Publishing the snapshot right after `ApplyChanges` closes the nil-snapshot gap. It reads the reconciled status and adds no new lock ordering.

What I verified:
- **The new test catches the bug.** `TestCoordinator_RestartPublishesRestoredAssignments` fails without the fix (`context deadline exceeded`) and passed 25/25 with `-race`.
- **The issue's full scenario is fixed.** Setup: RF=3, restart the coordinator while all replicas are healthy (no election), then restart a follower that was down about 12s.
  - Before this change, the follower stayed `NOT_SERVING` for 30s+.
  - With it, the follower was ready in about 1.4s.

Findings:

1. **Must fix: the release workflow still installs go-license `@latest`.** This PR pins go-license only in `pr_build_and_test.yaml`. `.github/workflows/release.yaml` (line 58) on this branch still runs `go install github.com/palantir/go-license@latest`.
   - CI runs with `GOTOOLCHAIN=local` on Go 1.26.8, and go-license v1.50.0 requires Go >= 1.27.0. The first CI run on this PR failed with exactly that error.
   - So the next v0.16.x tag will fail its build-test job. v0.16.10 was released on Sep 6, before v1.50.0 came out on Sep 9, which is why nobody has hit it yet.
   - Suggestion: mirror main's layout, where a single `GO_LICENSE_VERSION` env variable is used by both the build and release workflows, and set it to `v1.49.0` here.

2. **"Fixes #1292" covers the nil snapshot, but not the whole issue.** A data server that restarts faster than the coordinator's per-connection health ping (every 10s) still stays `NOT_SERVING` until the next assignment change.
   - The coordinator only ever `Send`s on the assignment stream. gRPC doesn't cancel a client stream's context when the server ends the stream, so the dispatcher keeps waiting on a dead stream. This is the case #1216 fixed on main.
   - It already happens on release-0.16 without any coordinator restart, so it isn't a regression from this PR.
   - Porting #1216's drain goroutine fixed it in a local test: the data server became ready in about 15s, bounded by the dispatch backoff.
   - I'd change this to "Part of #1292", then backport #1216 as a follow-up. That backport also needs a `RecvMsg` for the Maelstrom stub, since `BaseStream.RecvMsg` panics on 0.16 as well.

3. **Main first (see my comment above).**
   - On main, this is covered by the reconciler recomputing assignments at startup, plus #1215/#1216 on the stream side.
   - Main's coordinator has been refactored since, so a clean cherry-pick isn't possible and a 0.16-specific change is fine. Please reference the main changes it corresponds to in the description.
   - I couldn't find an equivalent of the new regression test on main. Please add it to main first (it should pass there), then carry it here.

4. **Behaviour change (fine, just FYI).** On a fresh cluster, the coordinator now publishes shards with an empty leader before the first election, so data servers become ready and `NewSyncClient` returns earlier.
   - The Go client gets `Unavailable` for the empty target and retries.
   - The same leaderless state already occurs during every election and on multi-shard startup, and it matches what main does.
   - The new wait in the control test covers the one test that relied on the old ordering. The other leader reads under `tests/` all wait for steady state or a successful `Put` first.

5. **Nit.** The regression test checks `WaitForNextUpdate` directly, which is deterministic. It doesn't exercise the data-server readiness path described in the issue, so an end-to-end readiness check would be a nice addition. Optional.

Unrelated to this PR: `TestCoordinator_ShrinkCluster` is flaky on release-0.16 with or without this change.
- The 0.16 copy of the test uses RF=3; the #1130 backport raised it from RF=1. So its node-removal swap races the follower catch-up after the first election, and a rejected swap is only retried on the 30s balancer tick, longer than the test's 10s wait.
- Main's version (#1199: RF=1, `Term >= 0`) passes reliably on 0.16.
- If CI goes red on it here, re-run.

mattisonchao:
>This is for the release-0.16 branch. In reality, I'd prefer all fixes to go into main first and only then get cherry-picked into maintenance branches.

The main already has similar logic. And we've got much different logic in 0.17 and 0.16.

PR #1307, opened 2026-09-19
chore: replace MPL magodo/slog2hclog with an in-repo hclog adapter

merlimat:
### Motivation

After Oxia's repositories were imported into the CNCF FOSSA instance, the license scan badge (added to the README in #1303) reads "failing". `github.com/magodo/slog2hclog` is one of the two reasons: it is MPL-2.0, which is not on the [CNCF allowlist](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md) and has no entry in [`license-exceptions/exceptions.json`](https://github.com/cncf/foundation/blob/main/license-exceptions/exceptions.json). It is a direct dependency of `oxiad`, used in one place: adapting our `*slog.Logger` to the `hclog.Logger` that hashicorp/raft logs to.

The other one, `juju/fslock` (LGPL-3.0), was replaced in #1306 (merged). The remaining MPL-2.0 dependencies (hashicorp/raft, go-immutable-radix, golang-lru, hcl) are covered by blanket "All CNCF Projects" exceptions and are left alone.

### Changes

- Add a small in-repo adapter, `oxiad/coordinator/metadata/provider/raft/raft_logger.go`, and use it in place of `slog2hclog.New(...)`. It depends on `github.com/hashicorp/go-hclog` (MIT), which becomes a direct dependency of `oxiad`.
- Raft only calls `Debug`, `Info`, `Warn`, `Error` and `With`. The rest of the `hclog.Logger` interface gets a minimal implementation rather than an embedded nil or null logger, so that a future raft release calling e.g. `Named()` neither panics nor silently drops its logs. Logger names are not supported.
- Behavior is unchanged: what gets emitted is decided by the slog handler, following its level even when it is reconfigured at runtime. As before, the level variable only backs `GetLevel()`/`SetLevel()`, which raft never calls.
- Remove `magodo/slog2hclog` from `go.mod`/`go.sum` in the three modules that list it (`oxiad` directly, `cmd` and `tests` indirectly). No other version moves and `go.work.sum` needs no change.

Why not a dependency: the only permissively licensed adapter in this direction, `ValerySidorin/shclog` (MIT), is a single untested v0.0.1 whose `Is*()` checks use `==` instead of a threshold, whose chained `With()` re-applies the accumulated attributes, and which snapshots the level at construction while Oxia reconfigures it at runtime. `evanphx/go-hclog-slog` adapts in the opposite direction, and go-hclog itself has no slog bridge.

### Verification

- The adapter tests for the subset raft uses (levels, handler-driven filtering, `With`, the level variable: 20 assertions) were first run against `slog2hclog` through a throwaway shim, then against the new adapter, with identical results.
- The log lines of a real raft node (`TestProvider/raft`) are identical before and after.
- `make license-check`, `make build` and `make lint` (CI-pinned golangci-lint v2.13.2, fresh cache) pass and leave the tree clean.
- `go test -race ./oxiad/coordinator/metadata/...` passes.
- `go-licenses csv ./...` from `cmd/` no longer lists `magodo/slog2hclog`, and reports `hashicorp/go-hclog` as MIT.

### Note for merging

#1306 has merged. It deleted the `oxiad/go.mod` line adjacent to the one this PR deletes, which conflicted as expected; `main` is now merged into this branch with that one-line conflict resolved (both lines deleted), and nothing else conflicted. With both changes present, `go-licenses csv ./...` from `cmd/` reports neither `juju/fslock` nor `magodo/slog2hclog`: the only entries outside the CNCF allowlist are the MPL-2.0 dependencies covered by the CNCF exceptions.


merlimat:
@copilot resolve the merge conflicts in this pull request

